PR #6, opened 2021-03-22
add dsnp message types

shannonwells:
Problem
=======
I needed to write these anyway in order to help us with our batch data generation and benchmarking, so I thought why not just add them to the SDK?

Solution
========
Added a dsnpTypes file with the types in TypeScript as described in the [DSNP Message Formats](https://libertydsnp.github.io/spec/Messages/Overview#dsnp-message-formats) part of the spec.


src/types/dsnpTypes.ts
```diff
@@ -0,0 +1,103 @@
...
+export interface Reaction {
+  fromAddress: EthereumAddress;
+  emoji: Uint8Array;
+  inReplyTo: Uint8Array;
+}
+
+// Private: an encrypted message of unknown type
+// fromAddress	id of the sender	bytes20
+// data	encrypted graph change data	string
+// messageID	keccak-256 hash of unencrypted content	bytes32
```
claireolmstead, line 98: This looks really good to me. I want to follow this comment pattern when we have to do this for the example client. Makes everything so much more clear.

claireolmstead (approved):

src/types/dsnpTypes.ts
shannonwells: I actually just grabbed the table text from the Spec :) made it easy to implement too.

src/types/dsnpTypes.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// Broadcast: a public post
+// fromAddress	ID of the sender	bytes20
+// messageID	keccak-256 hash of content stored at URI	bytes32
+// uri	content URI	string
+export interface DSNPBroadcast {
+  fromAddress: EthereumAddress;
+  messageID: Uint8Array;
+  uri: string;
+}
```
wilwade, line 11: We need these comments to match docblock style so they generate the docs nicely:

```suggestion
/**
 * Broadcast: a public post
 */
export interface DSNPBroadcast {
  /** ID of the sender */
  fromAddress: EthereumAddress;
  /** keccak-256 hash of content stored at URI */
  messageID: Uint8Array;
  /** content URI */
  uri: string;
}
```

src/types/dsnpTypes.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// Reply: a public reply post
+// fromAddress	ID of the sender	bytes20
+// inReplyTo	ID of the message the reply references	bytes32
+// messageID	keccak-256 hash of content stored at uri	bytes32
+// uri	content uri	string
+export interface DSNPReply {
```
wilwade, line 18: Should these all extend a type of `DSNPMessage` (or other name)? Or perhaps build up a union type of all of them?

src/types/dsnpTypes.ts
```diff
@@ -0,0 +1,103 @@
+import { EthereumAddress } from "@liberty30/lib-privacy-js";
```
wilwade, line 1: We will want to just pull this in instead. Also with TS 4.1 we can do template literal types! (https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)

```typescript
type EthereumAddress = `0x${string}`;
```

wilwade (changes requested):

src/types/dsnpTypes.ts
claireolmstead: Josh and I were talking about this and we thought that a "reply" was a comment on a post and "message" was a direct message.

shannonwells: I namespaced it.   I realized the specs aren't consistent in naming as well.  There is now a union type for DSNPMessage that has all of them.

shannonwells: @wilwade ^^

.github/workflows/main.yml
```diff
@@ -21,10 +21,10 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: '14.5.0'
+          node-version: '14.16.0'
 
       - name: Install
-        run: npm install
+        run: npm ci
```
shannonwells, line 27: I discovered this while trying to debug the CI failures. This is supposed to be faster and is recommended by GitHub.  It uses only the versions specified in package-lock.json.

.github/workflows/main.yml
```diff
@@ -21,10 +21,10 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: '14.5.0'
+          node-version: '14.15.5'
```
shannonwells, line 24: This is to make it match with what's in .node-version and .tool-versions.

src/types/Strings.d.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Special String Types
+ *
+ * @packageDocumentation
+ */
+
+/**
+ * 0x prefixed Hexadecimal String
+ */
+export declare type HexString = `0x${Lowercase<string>}`;
```
wilwade, line 10: I didn't know about `Lowercase`!

src/types/Strings.d.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * 0x prefixed Hexadecimal representation of an Ethereum Address
+ */
+export declare type EthereumAddress = HexString;
+
+/**
+ * 0x prefixed Hexadecimal representation of an Ethereum Contract that implements the Social Identity Interface
+ */
+export declare type SocialIdentityAddress = EthereumAddress;
```
wilwade, line 20: There are three types of ethereum addresses. All lower, all upper (equivalent) and mixed for https://eips.ethereum.org/EIPS/eip-55

Currently HexString only claims one, but we don't really have much control over which we are using.

src/testEnabler.test.ts
```diff
@@ -0,0 +1,40 @@
+import { DSNP, DSNPMessage } from "./types/DSNP";
```
wilwade, line 1: Should we rename this file `types.test.ts`?

wilwade (approved):
A few more questions, but I really like the namespace and moving to declaration files. Nothing super blocking, but check the comments.

PR #7, opened 2021-03-24
remove reference to private lib, remove Lowercase as it does not do what I thought it did

shannonwells:
Follow-up in answer to @wilwade 's questions

* removed a link which referred to one of our private libraries
* renamed the test file
* Lowercase didn't do what I thought it did, which was convert values automatically. It does not do type checking for capitalization or convert. It operates only on types, not even sure about the usefulness of it now.


enddynayn:
👍

wilwade (approved):

PR #8, opened 2021-03-31
Features/announcement module 177120830

aneyzberg:
Problem
=======
As a user of the SDK I want to be able to announce a stored batch.
[link to Pivotal Tracker #177120830](https://www.pivotaltracker.com/story/show/177120830)

Solution
========
We created a  new Module named Announce

with @rosalinekarr 

Change summary:
---------------
* Create an announcement module that successfully calls `batch` function on the `Announce` contract 
* Add integration tests
* Add documentation under `/docs/modules`
* Add utility called `activityPubHash` to ensure ActivityPub objects are ordered properly

Steps to Verify:
----------------
1. Spin up docker-node-poa
1. Deploy `Announce` contract from contracts repo to the docker-node-poa. 
1.  Retrieve contract address for `Announce` contract that was deployed from`Contracts` repo and set it as **BATCH_CONTRACT_ADDRESS** Env variable in .env file
1. In `.env` - Set RPC_URL - to the address a node on your  docker-node-poa network is running on
1. In `.env` - Set TESTING_PRIVATE_KEY to the private key of an account that exists on your docker-node-poa that has an account balance
1. run `npm run tests`
1. Integration test should run that successfully calls `batch` function on announce contract and receives correct result

TODO:
-------------
1. Ci is not yet configured and integration tests will not run on CI until [this story](https://www.pivotaltracker.com/story/show/177584391) is completed


enddynayn:
👍

src/announcement/announcement.test.ts
```diff
@@ -0,0 +1,54 @@
+//eslint-disable-next-line
+require("dotenv").config();
+import Web3 from "web3";
+import { keccak256 } from "js-sha3";
+
+import { batch } from "./announcement";
+
+const TESTING_PRIVATE_KEY = process.env.TESTING_PRIVATE_KEY as string;
+const RPC_URL = process.env.RPC_URL as string;
```
wilwade, line 9: Curious if `String(process.env.TESTING_PRIVATE_KEY)` is better than `as string`...

src/announcement/announcement.ts
```diff
@@ -0,0 +1,81 @@
...
+    stateMutability: "nonpayable",
+    type: "function",
+  },
+];
+
+const contract: Contract | null = null;
+
+export const getContract = (web3Instance: Web3): Contract => {
+  if (contract) return contract;
+  return new web3Instance.eth.Contract(BATCH_CONTRACT_ABI, BATCH_CONTRACT_ADDRESS);
```
wilwade, line 62: `BATCH_CONTRACT_ADDRESS` should likely be able to be passed in but default back to something. Any contract that matches the `IAnnounce` interface can be used.

src/announcement/announcement.ts
```diff
@@ -0,0 +1,81 @@
...
+    name: "batch",
+    outputs: [],
+    stateMutability: "nonpayable",
+    type: "function",
+  },
+];
+
+const contract: Contract | null = null;
+
+export const getContract = (web3Instance: Web3): Contract => {
```
wilwade, line 60: All exported methods should have a docblock. also we should have a lint that forces that...

src/announcement/announcement.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @param accountAddress   The address from which to post the batch
+ * @param uri       The URI of the hosted batch to post
+ * @param hash      A hash of the batch contents for use in verification
+ * @returns         A [web3 contract receipt promise](https://web3js.readthedocs.io/en/v1.3.4/web3-eth-contract.html#id36)
+ */
+export const batch = async ({ provider, accountAddress, uri, hash }: BatchArguments): Promise<any> => {
+  const contract = (await getContract(provider)) as Contract;
+  return await contract.methods.batch(hashPrefix(hash), uri).send({
+    from: accountAddress,
+    gas: 27000,
```
wilwade, line 79: I know why we hardcoded this, but we really don't want to hard code this. Do we want to switch to using estimate gas instead? (I guess we don't really want to do that in cases where the node is processing the transaction however. Hmm....)

src/types/hash.ts
```diff
@@ -0,0 +1 @@
+export type KeccakHash = string;
```
wilwade, line 1: I think this should also have a docblock

src/utilities/activityPubHash.ts
```diff
@@ -0,0 +1,33 @@
...
+ * activityPubHash() provides a simple way to hash activityPub objects while
+ * guaranteeing that identical objects with different key orders still return
+ * the same hash. The underlying hash method used is
+ * [Keccak256](https://en.wikipedia.org/wiki/SHA-3).
+ *
+ * @param data  The activityPub object to hashes
+ * @returns     A hexadecimal string containing the Keccak hash
+ */
+export const activityPubHash = (data: Record<string, unknown>): KeccakHash => {
+  const sortedData = sortJSON(data as Record<string, unknown>);
```
wilwade, line 30: I was imagining from the spec that the hash was happening upon the raw file, not the json. If we are going to change this to the json, we should update the spec.

Also sortJSON only sorts objects recursively, but not if there is an array of objects. Also I don't think JSON.parse or JSON.stringify guarantee order.

src/announcement/announcement.ts
```diff
@@ -0,0 +1,81 @@
...
+ * batch() allows users call the batch smart contract and post the URI and hash
+ * of a generated batch to the blockchain.
+ *
+ * @param provider  The web3 instance used for calling the smart contract
+ * @param accountAddress   The address from which to post the batch
+ * @param uri       The URI of the hosted batch to post
+ * @param hash      A hash of the batch contents for use in verification
+ * @returns         A [web3 contract receipt promise](https://web3js.readthedocs.io/en/v1.3.4/web3-eth-contract.html#id36)
+ */
+export const batch = async ({ provider, accountAddress, uri, hash }: BatchArguments): Promise<any> => {
```
wilwade, line 75: `any`

docs/modules/announcement_announcement.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Returns:** *Promise*<any\>
+
+A [web3 contract receipt promise](https://web3js.readthedocs.io/en/v1.3.4/web3-eth-contract.html#id36)
+
+___
+
+### getContract
+
+▸ `Const`**getContract**(`web3Instance`: *any*): *any*
```
wilwade, line 38: This is out of date.
I am the reason why we are merging these docs, but now that we are in a public repo, I think we should likely just setup a github pages for them and include the generation in the build step for packaging. I'll add some stories/notes for that.

wilwade (changes requested):

src/announcement/announcement.ts
rosalinekarr: This method shouldn't have been exported. Will fix.

src/announcement/announcement.ts
rosalinekarr: What would it default to? There isn't any one batch contract address that will work across everyone's docker nodes.

src/announcement/announcement.ts
rosalinekarr: `estimateGas` doesn't seem to work properly with POA nodes in the current version of web3.js. It returns a value exactly at the amount of gas needed which causes `send()` to throw `Signer Error: gas limit is too low. Need at least 25496`. We could use the estimate and add an extra `1000`, but then we'd be right back at having a constant in the method.

src/utilities/activityPubHash.ts
rosalinekarr: `JSON.stringify` stringifies in the order the keys are present in the object, and I don't think we want to sort arrays since they have a semantic order. Sorting the keys in records here is just to guarantee an identical hash.

src/announcement/announcement.ts
rosalinekarr: Unfortunately, `web3.js` does not expose the return type of contract send, so we don't have a type we can use here. We could change it to `Record<string, any>` or something like that, but it will be brittle and still have an `any` in it.

src/announcement/announcement.ts
wilwade: I'm thinking of @aneyzberg's work with testnet contract discovery. If the sdk knows what network the user is on, then it should also know what the default contract address is.

The sdk can know which network as the sdk will need to know what chain id. Chain Id will map to the network. (testnet vs betanet vs mainnet)

src/announcement/announcement.ts
wilwade: I think I prefer estimate + a constant small overage. The reason being is that different contract implementations may have different gas costs.

src/utilities/activityPubHash.ts
wilwade: JSON stringify does, but it isn't required to by the JSON spec.

Also this doesn't recursively sort objects in arrays:
```json
{
  attachments: [ {}, {} ]
}
```

Well ordered JSON is just not easy to guarantee due to the JSON spec.

src/announcement/announcement.ts
aneyzberg: @wilwade however doesn't this depend on the local network being deployed the exact same way every time? So that the address of the contract does not change

src/announcement/announcement.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const getContract = (web3Instance: Web3): Contract => {
+  if (contract) return contract;
+  return new web3Instance.eth.Contract(BATCH_CONTRACT_ABI, BATCH_CONTRACT_ADDRESS);
+};
+
+const getGasLimit = async (contract: Contract, uri: string, hash: KeccakHash, fromAddress: string): Promise<number> => {
+  const gasEstimate = await contract.methods.batch(hashPrefix(hash), uri).estimateGas({
+    from: fromAddress,
+    gas: 27000,
```
wilwade, line 61: Do we need gas here to run the estimate?

src/announcement/announcement.ts
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  return gasEstimate + 2000;
```
wilwade, line 64: 1. 2000 feels like a large buffer. Does it really need to be that large?
2. Can we put the 2000 into a constant?

src/announcement/announcement.ts
```diff
@@ -0,0 +1,85 @@
...
+ * batch() allows users call the batch smart contract and post the URI and hash
+ * of a generated batch to the blockchain.
+ *
+ * @param provider  The web3 instance used for calling the smart contract
+ * @param accountAddress   The address from which to post the batch
+ * @param uri       The URI of the hosted batch to post
+ * @param hash      A hash of the batch contents for use in verification
+ * @returns         A [web3 contract receipt promise](https://web3js.readthedocs.io/en/v1.3.4/web3-eth-contract.html#id36)
+ */
+export const batch = async (provider: Web3, accountAddress: string, uri: string, hash: KeccakHash): Promise<any> => {
```
wilwade, line 77: We need to use the https://github.com/ethereum-ts/Typechain that is produced by https://github.com/LibertyDSNP/contracts/. That will enable us to remove the any and properly type this return.

src/announcement/announcement.ts
wilwade: Not once we have our discovery code working.

aneyzberg: oh yeah. I suppose this is true.

src/utilities/activityPubHash.ts
rosalinekarr: So, it turns out this edge case gets sorted too, because arrays are technically subclasses of objects in javascript. An array like `const x = [1, 2, 3]` will get sorted by this method because `typeof x == "object"` will return true and `Object.keys(x)` will return `[0, 1, 2]`.

src/announcement/announcement.ts
aneyzberg: but I would say let's set a default once we actually have this discovery code working? Since currently setting a default would break things.

src/utilities/activityPubHash.ts
wilwade: Hmm.... Let's continue this conversation in the spec side. We can leave it for now, but we should write the spec and discuss it there.

wilwade: Added a spec story: https://www.pivotaltracker.com/story/show/177681451

wilwade (approved):

PR #10, opened 2021-04-05
Fix a couple of issues with the SDK and add some typescript types for DSNP messages

shannonwells:
Problem
=======
* The SDK's package name was invalid; capital letters are not allowed.
* The SDK needs some basic type definitions for us to build on; this was made immediate due to @enddynayn 's and my batch work
* Also why not update some packages too.
[Finishes #177633643](https://www.pivotaltracker.com/story/show/177633643)
[Related to #177164585](https://www.pivotaltracker.com/story/show/177164585)

Solution
========
* downcase the package name
* Add some type definitions, including DSNP message types and values.
* Update packages
* move jest configuration into its own .js file to make package.json less crowded

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@LibertyDSNP/sdk",
+  "name": "@liberty-dsnp/sdk",
   "version": "0.0.0",
   "description": "SDK Library for the DSNP",
   "main": "dist/main/index.js",
--- a/package.json
+++ b/package.json
@@ -35,21 +35,22 @@
     "node": ">=14.15.5"
   },
   "devDependencies": {
-    "ts-jest": "26.5.1",
-    "ts-node": "^9.1.1",
-    "typescript": "^4.1.5",
-    "@types/jest": "26.0.20",
-    "@typescript-eslint/eslint-plugin": "4.15.1",
-    "@typescript-eslint/parser": "4.15.1",
+    "@types/jest": "26.0.22",
+    "@typescript-eslint/eslint-plugin": "4.21.0",
+    "@typescript-eslint/parser": "4.21.0",
+    "eslint": "^7.23.0",
     "eslint-config-prettier": "7.2.0",
     "eslint-plugin-jest": "24.1.5",
     "eslint-plugin-prettier": "3.3.1",
-    "jest": "^26.6.3",
+    "jest": "26.6.3",
     "jest-junit": "12.0.0",
     "npm-run-all": "4.1.5",
     "prettier": "2.2.1",
+    "ts-jest": "26.5.4",
+    "ts-node": "9.1.1",
     "typedoc": "0.20.25",
-    "typedoc-plugin-markdown": "3.5.0"
+    "typedoc-plugin-markdown": "3.5.0",
+    "typescript": "4.2.3"
   },
   "jest": {
     "setupFiles": [
@@ -69,6 +70,5 @@
     "suiteNameTemplate": "{filepath}",
     "classNameTemplate": "{classname}",
     "titleTemplate": "{title}"
-  },
-  "dependencies": {}
+  }
 }
--- a/package.json
+++ b/package.json
@@ -32,14 +32,15 @@
     "*.js"
   ],
   "engines": {
-    "node": ">=14.5.0"
+    "node": ">=14.15.5"
   },
   "devDependencies": {
+    "ts-jest": "26.5.1",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.5",
     "@types/jest": "26.0.20",
     "@typescript-eslint/eslint-plugin": "4.15.1",
     "@typescript-eslint/parser": "4.15.1",
-    "babel-jest": "^26.6.3",
-    "eslint": "7.20.0",
     "eslint-config-prettier": "7.2.0",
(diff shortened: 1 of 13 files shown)


src/types/DSNP.d.ts
```diff
@@ -1,132 +1,137 @@
...
-  /**
-   * GraphChange: a public follow/unfollow
-   */
-  export interface GraphChange {
-    /** fromAddress	ID of the sender	bytes20 */
-    fromAddress: EthereumAddress;
-    /** actionType	follow/unfollow	number/enum */
-    actionType: number;
-  }
+/**
```
shannonwells, line 41: Bad diff! Bad!

src/types/DSNP.d.ts
```diff
@@ -1,132 +1,137 @@
 import { EthereumAddress } from "./Strings";
 
-export declare type DSNPMessage =
```
shannonwells, line 3: Turns out this doesn't work.  It is only ever a type and can only be used with type literals.

src/types/DSNP.d.ts
```diff
@@ -1,132 +1,137 @@
...
-   */
-  export interface Broadcast {
-    /** fromAddress	ID of the sender	bytes20 */
-    fromAddress: EthereumAddress;
-    /** messageID	keccak-256 hash of content stored at URI	bytes32 */
-    messageID: Uint8Array;
-    /** uri	content URI	string */
-    uri: string;
-  }
+export const ActionTypePrivate = 0;
```
shannonwells, line 5: I tried like mad to get namespaces to work and the advice I got from multiple sources was, give up on namespaces. So I do it here the "bonehead" way.

src/types/Strings.d.ts
```diff
@@ -5,17 +5,17 @@
  */
 
 /**
- * 0x prefixed Hexadecimal String
+ * Hexadecimal String
  */
-export declare type HexString = `0x${string}`;
```
shannonwells, line 10: This also doesn't work  the way I wanted; again, only works for string literals and that's not the purpose I wanted it for.

src/types/Strings.d.ts
```diff
@@ -5,17 +5,17 @@
...
+export declare type HexString = string;
 
 /**
- * 0x prefixed Hexadecimal representation of an Ethereum Address
```
shannonwells, line 13: Dropping this comment because it can't be enforced.

tsconfig.module.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "./tsconfig",
   "compilerOptions": {
-    "target": "es6",
```
shannonwells, line 4: redundant; these are overrides

src/types/index.js
```diff
@@ -0,0 +1,6 @@
+"use strict";
```
wilwade, line 1: This looks like a generated file.

src/types.test.ts
```diff
@@ -1,4 +1,4 @@
-import { DSNP, DSNPMessage } from "./types/DSNP";
+import * as DSNP from "./types/DSNP.d";
```
wilwade, line 1: Is the `.d` really needed for it to pick it up? It shouldn't be if I'm reading the module resolution docs correctly, but perhaps it depends on the type of module resolution we are using?
https://www.typescriptlang.org/docs/handbook/module-resolution.html#module-resolution-strategies

src/types/DSNP.d.ts
```diff
@@ -1,132 +1,137 @@
...
-   */
-  export interface Drop {
-    /** deadDropID	The Dead Drop ID (See DeadDrops	bytes32) */
-    deadDropID: Uint8Array;
-    /** messageID	keccak-256 hash of content	bytes32 */
-    messageID: Uint8Array;
-    /** uri	content uri	string */
-    uri: string;
-  }
+export type Message = {
```
wilwade, line 35: Should this be a type or an interface?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@LibertyDSNP/sdk",
+  "name": "@liberty-dsnp/sdk",
```
wilwade, line 2: Side note that this might change again as I think it will need to match our npm org which is going to be `unfinishedlabs` not our github org, but we can deal with that when we are publishing later this month.

wilwade (changes requested):
Nice. Few small things.

src/types.test.ts
shannonwells: I forgot about that; tsc was not finding the file.  This was a suggestion from tsc, but once I got the other things in place I forgot to take it out and try it.

src/types/index.js
shannonwells: right, good catch thanks

src/types/DSNP.d.ts
shannonwells: interface, thanks for catching it

wilwade (approved):
🎉

src/types.test.ts
wilwade: For future reference: We figured out that it was the `export const`s in the definition file that were the issue. Since .d.ts files are not compiled, those variables would disappear. Moved those out to a separate file and then we could remove the .d from the import and pull the types in.

PR #11, opened 2021-04-06
Features/stub method interfaces 177542925

rosalinekarr:
Problem
=======
We don't have a large architectural plan for modules and specific methods in the SDK, and that lack makes it difficult to start work on implementations when the interfaces they need to interact with are unclear.
[#177542925](https://www.pivotaltracker.com/story/show/177542925)

Solution
========
Add interfaces, stubs and documentation for as many modules and methods as we can think of for the v1 release. These interfaces and methods will likely change over the course of implementation, but their basic outline at least gives us a place to start and general concept of what other method calls will be available to us in implementation.

with @wilwade


src/utilities/hash.ts
```diff
@@ -0,0 +1,5 @@
+export type KeccakHash = string;
```
shannonwells, line 1: This type is exported in its own file already

src/types/hash.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * KeccakHash is a 256 bit hash string encoded as hex with a '0x' at the start
+ *
+ * @type KeccakHash
+ */
+export type KeccakHash = string;
```
shannonwells, line 6: If I understand how this is supposed to work correctly, I think we should follow putting this into type declarations i.e put them in a .d.ts file and do:
```
declare type KeccakHash = string
```

src/activityPub/activityPub.ts
```diff
@@ -0,0 +1,64 @@
+import { keccak256 } from "js-sha3";
+import { KeccakHash } from "../utilities/hash";
+import { NotImplementedError } from "../utilities/errors";
+
+const sortJSON = (obj: Record<string, unknown>): Record<string, unknown> => {
```
enddynayn, line 5: can we move this into utility file?

src/config/config.ts
```diff
@@ -0,0 +1,44 @@
+// import Web3 from "web3";
+// import MemoryStore from "../storage/memoryStorage";
```
enddynayn, line 2: maybe we can use node-convict for configuration?

src/utilities/hash.ts
rosalinekarr: Fixed

src/types/hash.ts
rosalinekarr: Fixed

src/activityPub/activityPub.ts
rosalinekarr: Yep, fixed

src/config/config.ts
rosalinekarr: sure, I think it's worth a research spike. would you mind adding a chore to the backlog?

wilwade: I took a quick look and I liked what I saw. Added it to the backlog to discuss https://www.pivotaltracker.com/story/show/177682091

enddynayn (approved):

src/config/config.ts
shannonwells: Same here, looks good at first glance

shannonwells (approved):
Just a couple of comments - thanks for doing this.  I agree I'm sure this isn't the final form so I think this is a great start.

src/batch/batch.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @returns         A batch file object
+ */
+export const createFile = (_events: MessageType[]): BatchFileObject => {
+  throw NotImplementedError;
+};
+
+/**
+ * openFile() takes a Batch file object . This method is not yet implemented.
+ *
+ * @params  file  An array of DSNPEvent to include in the batch file
```
aneyzberg, line 20: is this description correct? It seems like a copy and past error from above

src/batch/batch.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @returns       A batch file object
```
aneyzberg, line 21: It seems like the return value might be incorrect here based on the message signature below?

aneyzberg (approved):
lgtm just a couple small comments

src/contracts/announcement.ts
```diff
@@ -4,8 +4,8 @@ import Web3 from "web3";
 import { Contract } from "web3-eth-contract";
 import { AbiItem } from "web3-utils";
 
-import { KeccakHash } from "../types/hash";
-import { hashPrefix } from "../utilities/hashPrefix";
+import { HexString } from "../types/Strings";
```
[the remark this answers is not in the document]
rosalinekarr: Not sure, but I think we can save making a decision in a future PR.

src/batch/batch.ts
rosalinekarr: fixed

src/batch/batch.ts
rosalinekarr: fixed

PR #21, opened 2021-05-13
Features/events module 177859533

rosalinekarr:
Problem
=======
We need a way to create DSNP messages in the SDK for later signing and inclusion in batches.
[#177859533](https://www.pivotaltracker.com/story/show/177859533)

Solution
========
Added a messages module with factories and types for Broadcast, Reply and Reaction events.

Change summary:
---------------
* Added types for Broadcast, Reply and Reaction events
* Added factory functions for each type

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,10917 @@
 {
   "name": "@unfinishedlabs/sdk",
   "version": "0.0.0",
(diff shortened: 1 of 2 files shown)


enddynayn:
fantastic!

enddynayn (approved):

PR #22, opened 2021-05-13
feat(announcement): add read methods

enddynayn:
Problem
=======
problem statement, including
[link to Pivotal Tracker #12345678](https://www.pivotaltracker.com/story/show/12345678)

Solution
========
What I/we did to solve this problem

with @pairperson1

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing


src/batch/batch.ts
```diff
@@ -1,25 +1,83 @@
...
- * @params  file  A batch file object to open
- * @returns       An array of DSNPEvents from the batch file
+ * @param  reader a ParquetReader object.
+ * @param  column a ParquetReader object.
+ * @param  item a ParquetReader object.
+ * @returns void.
  */
-export const openFile = (_file: BatchFileObject): MessageType[] => {
-  throw NotImplementedError;
+export const batchIncludes = async (
```
enddynayn, line 75: write test

shannonwells (changes requested):
The code looks clean and straightforward.

shannonwells (approved):

src/batch/batch.ts
```diff
@@ -1,25 +1,85 @@
 import { MessageType } from "../types/DSNP";
 import { NotImplementedError } from "../utilities";
+import request from "request";
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const parquet = require("@unfinishedlabs/parquetjs");
```
wilwade, line 5: Does this not work?
```suggestion
import * as parquet from "@unfinishedlabs/parquetjs";
```

src/batch/batch.ts
```diff
@@ -1,25 +1,85 @@
...
+ * readFile() reads a Parquet file by row.
  *
- * @params  file  A batch file object to open
- * @returns       An array of DSNPEvents from the batch file
+ * @param reader a ParquetReader object.
+ * @returns void.
  */
-export const openFile = (_file: BatchFileObject): MessageType[] => {
-  throw NotImplementedError;
+export const readFile = async (reader: typeof ParquetReader, callback: ReadRowFunction): Promise<void> => {
```
wilwade, line 57: It is more complex than a callback function, but do we want to have this be an Iterable?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators

src/batch/batch.test.ts
```diff
@@ -0,0 +1,32 @@
+import { batchIncludes } from "./batch";
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { ParquetReader } = require("@unfinishedlabs/parquetjs");
```
wilwade, line 4: Does this not work?

```suggestion
import { ParquetReader } from "@unfinishedlabs/parquetjs"
```

wilwade (commented):
Two questions and a wonder about Iterable

src/batch/batch.ts
enddynayn: it does not

src/batch/batch.ts
enddynayn: the reader is already iterable.

aneyzberg (approved):
lgtm

PR #23, opened 2021-05-14
Features/message signing 177926856

rosalinekarr:
Problem
=======
We need a way to sign messages for announcement to prove authenticity.
[#177926856](https://www.pivotaltracker.com/story/show/177926856)

Solution
========
Add `#sign()`, `#verify()` and `#hash()` methods to the message module


src/messages/messages.ts
```diff
@@ -99,3 +103,45 @@ export const createReactionMessage = (fromId: string, emoji: string, inReplyTo:
...
+ * sign() takes a private key and DSNP message and returns a Uint8Array
+ * signature for the message.
+ *
+ * @param privateKey The private key to use for signing
+ * @param message    The DSNP message to sign
+ * @returns          The message signature as a Uint8Array
+ */
+export const sign = (privateKey: Uint8Array, message: DSNPMessage): Uint8Array =>
+  secp256k1.ecdsaSign(hash(message), privateKey).signature;
+
```
wilwade, line 136: This feels like a wallet problem, not an SDK problem. I know we are skirting around that issue. Perhaps we need both?

wilwade (commented):
Question re: wallets

shannonwells (approved):

aneyzberg (approved):

src/messages/messages.ts
```diff
@@ -99,3 +102,48 @@ export const createReactionMessage = (fromId: string, emoji: string, inReplyTo:
...
+/**
+ * verify() takes a public key, DSNP message and a message signature and returns
+ * whether the signature is valid for the given key and message.
+ *
+ * @param message   The DSNP message to sign
+ * @param signature The message signature to validate
+ * @returns         The address of the signer
+ */
+export const verify = (message: DSNPMessage, signature: string): string => {
+  return ethers.utils.verifyMessage(hash(message), signature);
```
wilwade, line 148: I think we want to use `ethers.utils.recoverPublicKey` here to prevent using all the special Ethereum stuff we don't need.

wilwade (changes requested):

rosalinekarr:
This PR is currently blocked by this spec change: https://github.com/LibertyDSNP/spec/pull/68

src/messages/messages.ts
rosalinekarr: **Notes following out-of-band conversation:**

- There's no way around the Ethereum's RPC prefix
- We have to update the spec to use the prefix instead
- PR to do so here: https://github.com/LibertyDSNP/spec/pull/68

src/messages/messages.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        ...(await getConfig()),
+        signer: new ethers.Wallet(privateKey),
+      });
+    });
+
+    it("returns a valid signature for a valid private key and message", async () => {
+      const message = createBroadcastMessage("1", "https://dsnp.org", "0x12345");
+
+      expect(await sign(message)).toEqual(
+        "0x9bde6e79c896768bfeed17e6fb7508d4caca83d01a18e5d14a4428e45d8aced33851595e4234e0d4396a8459109e1eb284583f6745a64289723e41aba1b92a8d1c"
```
acruikshank, line 53: I see that there is effectively a round trip test here, but it'd be nice to actually see the round trip. Confirming that the signature equals this long fixture doesn't provide much value. I'd be nice to see that signing a message, and then verifying with the same signature and a different message produces the wrong public key.

acruikshank (changes requested):
Hashing the message before calling signMessage (which hashes it again) is unnecessary and goes against the spec. Other than that and the verify/recover issue, it's looking good.

src/messages/messages.ts
```diff
@@ -99,3 +102,48 @@ export const createReactionMessage = (fromId: string, emoji: string, inReplyTo:
...
+ * signature for the message.
+ *
+ * @param message The DSNP message to sign
+ * @param opts    Optional. Configuration overrides, such as from address, if any
+ * @returns       The message signature as a Uint8Array
+ */
+export const sign = async (message: DSNPMessage, opts?: Config): Promise<HexString> => {
+  const { signer } = await getConfig(opts);
+  if (!signer) throw MissingSigner;
+  return signer.signMessage(hash(message));
```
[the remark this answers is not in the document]
rosalinekarr: yep, see [prior conversation](https://github.com/LibertyDSNP/sdk-ts/pull/23#discussion_r636932541)

src/messages/messages.ts
```diff
@@ -99,3 +102,48 @@ export const createReactionMessage = (fromId: string, emoji: string, inReplyTo:
...
+};
+
+/**
+ * verify() takes a public key, DSNP message and a message signature and returns
+ * whether the signature is valid for the given key and message.
```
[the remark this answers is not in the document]
rosalinekarr: updated

src/messages/messages.test.ts
rosalinekarr: These tests use the same public-private key pair and message, so in effect, they already are a round-trip test.

rosalinekarr: Added one anyway: https://github.com/LibertyDSNP/sdk-ts/pull/23/files#diff-a7f6db83f43ab14cc6a3c03d3eeb8f80dd905b6c0e01685ade183f25ee5f5b73R46

src/messages/messages.ts
```diff
@@ -99,3 +101,43 @@ export const createReactionMessage = (fromId: string, emoji: string, inReplyTo:
...
+ * @throws {@link MissingSigner}
+ * This error is thrown if no Signer is defined in the configuration options.
+ *
+ * @param message The DSNP message to sign
+ * @param opts    Optional. Configuration overrides, such as from address, if any
+ * @returns       The message signature in hex
+ */
+export const sign = async (message: DSNPMessage, opts?: Config): Promise<HexString> => {
+  const { signer } = await getConfig(opts);
+  if (!signer) throw MissingSigner;
```
wilwade, line 129: Side note, we should likely be throwing this in the config module instead. Perhaps `config.getSigner(opts)`, but I think that is a separate story.

wilwade (approved):
🚢

PR #25, opened 2021-05-14
fix underscore vuln

shannonwells:
Problem
=======
The version of web3 we were using, plus what we had in package-lock.json was using a version of underscore that had a severe vulnerability.   

Also there is a severe vulnerability in lodash.

[link to Pivotal Tracker #178101352](https://www.pivotaltracker.com/story/show/178101352)

Solution
========
1. Installed latest versions of underscore and lodash explicitly
1. Completely regenerate package-lock.json
1. Run npm audit fix

This leaves us only with this ["low" web3 vulnerability](https://www.npmjs.com/advisories/877) 

Steps to Verify:
----------------
1. `npm i`
2. `npm audit` if this does not show you the remaining vulnerability. Only the 1 should be there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "js-sha3": "^0.8.0",
-        "lodash": "^4.17.21",
         "typescript": "^4.1.5",
-        "underscore": "^1.13.1",
         "web3": "^1.3.5 ",
         "web3-eth-contract": "^1.3.5"
       },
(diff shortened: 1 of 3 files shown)


enddynayn:
do we need to use both underscore and lodash?

wilwade:
> do we need to use both underscore and lodash?

I think @enddynayn is correct. Now that the package-lock.json file has the newer versions, we can remove them from the dependencies list. They are just sub dependencies, so we just needed the lock file to reflect the updates.

package-lock.json
```diff
@@ -11,7 +11,9 @@
       "license": "Apache-2.0",
       "dependencies": {
         "js-sha3": "^0.8.0",
+        "lodash": "^4.17.21",
         "typescript": "^4.1.5",
+        "underscore": "^1.13.1",
```
wilwade, line 16: If I run `npm i` npm edits the package-lock file, but after that change, it does correctly return clear for `npm audit`. Can you check if your `npm i` triggers any additional changes?

wilwade (commented):
One question

package-lock.json
wilwade: Also a rebase should fix all the test issues

wilwade (approved):
🔒

PR #27, opened 2021-05-18
add parquet-schema

shannonwells:
Problem
=======
To write batch files we will need Parquet schema for every DSNP Announcement type.
[Finishes #177940101](https://www.pivotaltracker.com/story/show/177940101)

TODO: review the schema and make sure they are up to date with the message fields.

Solution
========
Add the schema files

with @pairperson1

### DIFF
--- a/src/batch/parquet-schema.ts
+++ b/src/batch/parquet-schema.ts
@@ -8,24 +8,29 @@
  */
 export const BroadcastSchema = {
   actionType: { type: "INT32" },
+  contentHash: { type: "BYTE_ARRAY" },
   fromAddress: { type: "BYTE_ARRAY" },
-  messageID: { type: "BYTE_ARRAY" },
   uri: { type: "BYTE_ARRAY" },
 };
 
+/**
+ * Reply: A public reply post
+ */
 export const ReplySchema = {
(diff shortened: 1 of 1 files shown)


enddynayn:
looking great!

enddynayn:
amazing!

enddynayn (approved):

src/batch/parquet-schema.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Broadcast: a public post
+ */
+export const BroadcastSchema = {
+  actionType: { type: "INT32" },
+  fromAddress: { type: "BYTE_ARRAY" },
+  messageID: { type: "BYTE_ARRAY" },
+  uri: { type: "BYTE_ARRAY" },
+};
+
+export const ReplySchema = {
```
aneyzberg, line 16: the other schemas have comments explaining what they are for. Do we want a comment for reply schema and drop schema below?

aneyzberg (approved):
small comment but otherwise lgtm

src/batch/parquet-schema.ts
shannonwells: Yes, I added them.

PR #30, opened 2021-05-18
Chore/switch to ethers 178091932

acruikshank:
Problem
=======
We believe Ethers.js is better framework for working with Etherium nodes than Web3.js. We also use Ethers.js for contract development (in our tests). The SDK has been built on Web3.js.
[link to Pivotal Tracker #178091932](https://www.pivotaltracker.com/story/show/178091932)

Solution
========
 This PR re-implements all existing functionality and tests with Ethers.js

Change summary:
---------------
* Build ethers-v5 Typescript classes instead of web3-v1.
* Add provider and signer to config rather than (web3) provider and account address.
* Rewrite Announcer, Contract and their tests to use Ethers.js.

Steps to Verify:
----------------
1. Run the code. Run the tests. Should work as before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,10 +12,10 @@
   "scripts": {
     "build": "npm run clean && run-p build:*",
     "build:cjs": "tsc -p tsconfig.json",
-    "build:types": "tsc -p tsconfig.types.json && mkdir -p ./dist/types/types && cp ./src/types/*.d.ts ./dist/types/types",
+    "build:types": "tsc -p tsconfig.types.json && mkdir -p ./dist/types/types && rsync -a --filter='+ */' --filter='-! **/*.d.ts' ./src/types/* ./dist/types/types",
     "build:esm": "tsc -p tsconfig.esm.json && echo '{\"type\": \"module\"}'> dist/esm/package.json",
     "build:multimodules": "node scripts/multimodules.js",
-    "coverage": "jest test --coverage --collectCoverageFrom='src/**/*.ts'",
+    "coverage": "jest test --runInBand --coverage --collectCoverageFrom='src/**/*.ts'",
     "clean": "rm -Rf dist docs ./*.d.ts ./*.js",
     "doc": "rm -Rf dist/docs && run-s doc:html",
     "doc:html": "typedoc src/ --plugin none --theme default --out dist/docs",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -94,6 +95,7 @@
     "titleTemplate": "{title}"
   },
   "dependencies": {
+    "ethers": "^5.1.4",
     "js-sha3": "^0.8.0",
     "typescript": "^4.1.5",
     "web3": "^1.3.5 ",
```
rosalinekarr, line 101: I thought we were removing `web3` entirely. Can we drop this dependency and the one after it?

src/config/config.ts
```diff
@@ -1,14 +1,13 @@
...
 import { QueueInterface } from "../batch/queue";
 import MemoryStore from "../storage/memoryStorage";
 import { StorageInterface } from "../storage/storage";
-import { EthereumAddress } from "../types/Strings";
 
 export interface Config {
-  accountAddress?: EthereumAddress;
-  provider?: Web3;
+  provider?: ethers.providers.Provider;
+  signer?: ethers.Signer;
```
rosalinekarr, line 10: It doesn't necessarily need to happen in this PR, but now that we're using ethers, I wonder if we can remove the optionality from these keys and include a default ethers provider in the init below.

rosalinekarr (changes requested):
one blocking item and one non-blocking, otherwise looks good

package.json
acruikshank: good point.

rosalinekarr (approved):

src/test/hardhatRPC.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe("snapshot and revert", () => {
+  it("clears changes after revert", async () => {
+    jest.setTimeout(12000);
+
+    // snapshot
+    await snapshotHardhat(provider);
+
+    // create a batch
+    await batch("0x" + keccak256("test"), "http://www.testconst.com");
```
enddynayn, line 31: hmm I think perhaps we need a rebase.

enddynayn:
excellent!

enddynayn (approved):

aneyzberg (approved):
lgtm minus the need to rebase

PR #31, opened 2021-05-19
Fix sdk-ts failing postinstall b/c typechain is not found

shannonwells:
Problem
=======
When I tried to add the sdk to the test-generators repo, it failed because test-generators does not depend on typechain, but typechain is called by the postinstall script for this package.  I believe moving the typechain from devDependencies to dependencies [Fixes #178206018](https://www.pivotaltracker.com/story/show/178206018).  When I did this locally the test-generators package-json included typechain in the sub-dependencies and I was able to install it by referring to the sdk project folder after building.

Solution
========
1. move typechain to dependencies
1. Also:  bump nodejs in .tool-versions 

Steps to Verify:
----------------
You should be able to install this package in a project without having typechain part of the current project.


enddynayn (approved):

wilwade (approved):
👍

enddynayn (approved):

PR #34, opened 2021-05-20
rename messageID to contentHash

aneyzberg:
Problem
=======
Rename `messageID` to `contentHash` per 
[link to Pivotal Tracker #178202728](https://www.pivotaltracker.com/story/show/178202728)

### DIFF
--- a/src/types/DSNP.d.ts
+++ b/src/types/DSNP.d.ts
@@ -25,8 +25,8 @@ export interface Message {
  * Broadcast: a public post
  */
 export interface Broadcast extends Message {
-  /** messageID keccak-256 hash of content stored at URI bytes32 */
-  messageID: string;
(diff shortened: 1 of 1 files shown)


src/types/DSNP.d.ts
```diff
@@ -25,8 +25,8 @@ export interface Message {
  * Broadcast: a public post
  */
 export interface Broadcast extends Message {
```
wilwade, line 27: Wait a second. This is a duplicate of https://github.com/LibertyDSNP/sdk-ts/blob/main/src/messages/messages.ts#L22

How do we have two of these.

wilwade (changes requested):
We have some duplicate types to reconcile.

aneyzberg:
Closing as this file is no longer needed

PR #35, opened 2021-05-22
Some test data generation code

shannonwells:
Problem
=======
We need some ways of generating test data fixtures for use in general testing as well as things like benchmarks.
[Part of #178254168](https://www.pivotaltracker.com/story/show/178254168)

1. Add a generator that generates DSNP Message events
2. Add some generators that generate ActivityPubs
3. Add some generators that generate and write parquet batch file fixtures.

Also:
Fix bugs discovered while writing tests for the above. Yay tests!

### TODO
- [x] wait for the test-generators PR to be merged and use that instead of the linked version
- [x] add stuff to Bloom filters so we generate files like we're expecting for DSNP
- [x] make sure tests all pass
- [x] any final rebasing
- [x] make sure CI passes


enddynayn (approved):

src/test/generators/activityPubGenerators.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+};
+
+/**
+ * Generate a Profile update. The type `Person` is not a `Profile`
+ * @param address the HexString socialAddress to make this person update around
+ * @param name (optional) the new name of the profile update, or generate one.
+ * @param username (optional) the new username of the profile update TODO
+ */
+export const generatePerson = (address: HexString, name?: string): ActivityPub => {
```
enddynayn, line 44: good to know!

src/batch/parquet-schema.ts
```diff
@@ -14,18 +28,26 @@ export const BroadcastSchema = {
   signature: { type: "BYTE_ARRAY" },
 };
 
+export const BroadcastBloomFilterOptions: BloomFilterOptions = {
```
shannonwells, line 31: Note here in all these bloom filter options, you can opt to add any extras; this const and the others simple ensures we have the ones we want, i.e. standardizes on the columns to add to the bloom filter,  based on the spec.

enddynayn:
👍 beautiful!

enddynayn (approved):

PR #38, opened 2021-05-26
refact(store): update storage

enddynayn:
Update storage interface and methods.


src/storage/storage.ts
```diff
@@ -1,46 +1,44 @@
...
+ * @param   _targetPath  The URI of the batch file to download
  * @returns       The batch file fetched from the given URI
  *
  * @throws HTTPError
  * Indicates an error was thrown at the HTTP layer. This may indicate that the
  * host has censored the content for legal reasons if the error code is 451.
  */
-export const fetch = async (_uri: string, _opts?: Config): Promise<File> => {
-  throw NotImplementedError;
+export const get = async (_targetPath: string): Promise<File> => {
```
wilwade, line 39: Why the underscore?

src/storage/storage.ts
```diff
@@ -1,46 +1,44 @@
...
 
-export type File = BatchFileObject | ActivityPub;
+export type File = Buffer | string;
+export type Content = string | Buffer;
 
 export interface StorageInterface {
-  store(file: File): Promise<string>;
-  fetch(uri: string): Promise<File>;
+  put(targetPath: string, content: Content): Promise<URL>;
+  get(targetPath: string): Promise<File>;
```
wilwade, line 9: This is not really related to this I wonder how we are going to use get? I like that fetch is gone. URI fetching should not happen here. I guess this could be used for recovering only the files you stored via put.

src/config/config.ts
```diff
@@ -23,7 +21,7 @@ let config: Config = {
  *
  * @returns The current configuration settings
  */
-export const getConfig = async (overrides?: Config): Promise<Config> => {
+export const getConfig = (overrides?: Config): Config => {
```
wilwade, line 24: 🎉

wilwade (commented):
Two questions, but it looks great!

src/storage/storage.ts
enddynayn: I was not sure why this was underscored myself. I just kept it as is. However, now that you point it out I will update it.

src/storage/storage.ts
shannonwells: It can also be used by archivers for storage, and indexers who want to get new batches and read/parse from local storage.

src/config/config.ts
```diff
@@ -2,19 +2,17 @@ import { ethers } from "ethers";
...
 import { QueueInterface } from "../batch/queue";
-import MemoryStore from "../storage/memoryStorage";
 import { StorageInterface } from "../storage/storage";
 
 export interface Config {
   provider?: ethers.providers.Provider;
   signer?: ethers.Signer;
   queue: QueueInterface;
-  store: StorageInterface;
+  store?: StorageInterface;
```
shannonwells, line 11: Is this meant to be optional?  It throws in get & put if there's no store in the config.

shannonwells (changes requested):
I'm not sure what your vision is for how this is to be used. Seems like there needs to be a default store if the config value is optional.  Or do you expect that setConfig with a storage interface will be called?

If you had a couple of tests even if they were very short and only stubbed out the StorageInterface I  think this would be clearer.

src/storage/storage.ts
enddynayn: Yeah, I agree with this. I asked this question to Shannon yesterday. But left it just in case I missed something.

src/config/config.ts
enddynayn: Yeah I was going to do that in the next PR. I wanted to keep this as simple as possible.

src/storage/storage.ts
enddynayn: looks like Shannon answered question.

src/config/config.ts
shannonwells: oh I see ok all that makes sense, thanks.

shannonwells (approved):
Approved given the planned follow-up

PR #40, opened 2021-05-29
Features/queue module 178334037

rosalinekarr:
Problem
=======
We haven't yet implemented the queue module for storing DSNP messages to be announced.
[#178334037](https://www.pivotaltracker.com/story/show/178334037)

Solution
========
Implement it

Change summary:
---------------
* Moved queue files to a separate directory from `batch`
* Uncommented and updated `src/queue/queue.ts`
* Implemented MemoryQueue for in-memory queuing
* Added tests


aneyzberg (approved):
small comment. but lgtm

src/queue/queue.ts
```diff
@@ -0,0 +1,57 @@
+// import * as batch from "../batch/batch";
+import { getConfig, Config } from "../config/config";
+// import * as announcement from "../contracts/announcement";
```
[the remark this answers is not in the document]
rosalinekarr: yeah, they'll be used once [`commit()`](https://github.com/LibertyDSNP/sdk-ts/pull/40/files/5442ca8228f04fdf35702fd414d5a8f85157dfe4#diff-9293f95797e2cb395a23e30cd19778715f9b6bc64dbbfd6c4f384b6680bb037bR47) is implemented below, but that doesn't need to happen until [this task](https://www.pivotaltracker.com/story/show/178272032).

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,65 @@
+import { QueueId, QueueInterface } from "./queue";
+import { DSNPMessage } from "../messages/messages";
+
+export const IdDoesNotExist = new Error("No message matching the given ID exists.");
```
wilwade, line 4: Should this be exported?

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * MemoryQueue provides a simple, in-memory store for DSNP messages to be
+ * announced. This adapter is the default if no other is provided in
+ * configuration.
+ */
+export default class MemoryQueue implements QueueInterface {
```
wilwade, line 11: Just want to double check that we don't want to make this a module instead of a class.

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * getAll() implements the getAll function of the QueueInterface. It returns an
+   * array of all items currently in the queue and clears the queue.
+   *
+   * @returns An array of all messages currently in the queue
+   */
+  async getAll(): Promise<DSNPMessage[]> {
+    const messages = this.queue.filter((msg) => msg !== undefined) as DSNPMessage[];
```
wilwade, line 59: This as DSNPMessage[] can be avoided with a guard (now called type predicates apparently) https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/queue/queue.ts
```diff
@@ -0,0 +1,57 @@
...
+  throw NotImplementedError;
+
+  // const config = getConfig(opts);
+  // const events = await config.queue.getAll();
+  // const batchFile = batch.createFile(events);
+  //
+  // const hash = batch.hash(batchFile);
+  // const uri = await storage.put(batchFile, opts);
+  // await announcement.batch(uri, hash);
+};
```
wilwade, line 57: Not related to this in particular, but I wonder if this shouldn't just have a `popAll` instead of a commit. I don't like the idea that this would call the batch module or that the batch module would call this one. That feels higher level than the module.

src/queue/memoryQueue.ts
rosalinekarr: It's used in tests, but I don't have any other good reason to export it.

I suppose a user might want to try-catch the `dequeue` method and ignore this error in particular while still raising exceptions for memory issues, but they could also accomplish that by just implementing their own queue.

src/queue/memoryQueue.ts
rosalinekarr: I made it a module at first, but the class version is preferable for resetting the queue in memory. As a module, dumping the entire queue and starting over from empty is a pain because it requires re-importing.

src/queue/queue.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * commit() creates a batch file from the current activity pub messages in the
+ * queue then clears the queue. This method is not yet implemented.
+ */
+export const commit = async (_opts?: Config): Promise<void> => {
+  throw NotImplementedError;
+
+  // const config = getConfig(opts);
+  // const events = await config.queue.getAll();
+  // const batchFile = batch.createFile(events);
```
enddynayn, line 52: this would create multiple batch files right? As a result would change the rest of this method.

src/queue/queue.ts
rosalinekarr: Good idea. I'll add a note to the story for implementing this.

src/queue/queue.ts
enddynayn: Also, did we decide to keep all different type of messages in one queue?

src/queue/memoryQueue.ts
wilwade: Wouldn't we want to have custom error classes then?

src/queue/memoryQueue.ts
wilwade: ... I don't quite understand the response. Resetting the queue would need to happen in the same way for the class version no? You cannot just `setConfig({queue: new Queue()})` because after you committed something else might still be holding on to a reference of the old queue.

src/queue/queue.ts
```diff
@@ -0,0 +1,57 @@
...
+import { DSNPMessage } from "../messages/messages";
+import { HexString } from "../types/Strings";
+import { NotImplementedError } from "../utilities";
+
+export type QueueId = HexString;
+
+export interface QueueInterface {
+  enqueue(dsnpMessage: DSNPMessage): Promise<QueueId>;
+  dequeue(id: QueueId): Promise<DSNPMessage>;
+  getAll(): Promise<DSNPMessage[]>;
```
enddynayn, line 14: peak?

src/queue/queue.ts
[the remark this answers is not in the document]
rosalinekarr: removed these after scraping the `commit()` method in favor of `dequeueBatch()` at @wilwade's [suggestion](https://github.com/LibertyDSNP/sdk-ts/pull/40#discussion_r643427787)

src/queue/memoryQueue.ts
rosalinekarr: Removed the export

src/queue/memoryQueue.ts
rosalinekarr: No longer relevant after last re-write

src/queue/queue.ts
rosalinekarr: Replaced `commit()` with a `dequeueBatch()` function. I chose `dequeueBatch` over `popBatch` because the function pulls items from the front of the queue rather than the end like a stack, and I chose `dequeueBatch` over `dequeueAll` since it takes a `count` parameter and doesn't necessarily dequeue the entire list.

src/queue/queue.ts
rosalinekarr: We did not decide that, and that was a mistake on my part. I've re-written the entire queue to store messages of different DSNPTypes separately now.

src/queue/queue.ts
rosalinekarr: I scrapped `getAll()` in the latest re-write. `peek()` might be a good addition, but at the moment, I'm trying to keep the interface as minimal as possible as to not accidentally exclude any future queue adapters.

package-lock.json
```diff
@@ -12,12 +12,12 @@
       "dependencies": {
         "@dsnp/parquetjs": "^0.0.0-2ccc4e",
         "@ethersproject/abi": "^5.1.2",
-        "ethers": "^5.3.0",
+        "ethers": "^5.1.4",
```
wilwade, line 15: Something keeps bumping ethers down.

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * enqueue() implements the enqueue function of the QueueInterface. It takes a
+   * DSNP message, adds it to an in-memory store and returns an id which can be
+   * used to reverse this action, if needed.
+   *
+   * @param message The DSNP message to queue up for the next batch
+   * @returns       A queue id to be used to remove the message if needed
+   */
+  async enqueue(type: DSNPType, message: DSNPMessage): Promise<QueueId> {
```
wilwade, line 34: Doesn't a DSNPMessage contain the type?

src/queue/queue.ts
```diff
@@ -0,0 +1,59 @@
+import { getConfig, Config } from "../config/config";
+import { DSNPMessage, DSNPType } from "../messages/messages";
+
+export type QueueId = string;
+
+export interface QueueInterface {
```
wilwade, line 6: These exports should have docs

wilwade (changes requested):
Two small things. Looking nice!

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,105 @@
+import { QueueId, QueueInterface } from "./queue";
```
enddynayn, line 1: nice!

enddynayn (approved):

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * dequeue() implements the dequeue function of the QueueInterface. It takes a
+   * DSNP type, then removes and returns the first DSNP message of the matching
+   * type added to the queue. If no messages of the type exist, null will be
+   * returned to indicate the end of the queue.
+   *
+   * @param type The DSNP message type to dequeue
+   * @returns    The dequeued DSNP message or null to indicate end of queue
+   */
+  async dequeue(type: DSNPType): Promise<DSNPMessage | null> {
```
wilwade, line 54: Something you mentioned in standup made me wonder. If dequeue takes a type, how do I know what all the types are to check for?

src/queue/memoryQueue.ts
rosalinekarr: It does, but I was thinking pulling the type out would make implementation easier for future storage adapters like Redis. I'll change it.

src/queue/memoryQueue.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const index = this.queues[typeString].findIndex((queueItem) => queueItem.id === id);
+    if (index === -1) throw IdDoesNotExist;
+
+    const msg = this.queues[typeString][index].message;
+    this.queues[typeString].splice(index, 1);
+    return msg;
+  }
+
+  _createQueueItem(message: DSNPMessage): QueueItem {
```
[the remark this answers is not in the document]
rosalinekarr: It's [a convention for private methods](https://www.robinwieruch.de/javascript-naming-conventions#javascript-naming-conventions-private) although it doesn't actually make anything private.

There is a future [proposal for javascript private methods](https://caniuse.com/mdn-javascript_classes_private_class_fields), but unfortunately it's not well supported yet.

src/queue/memoryQueue.ts
rosalinekarr: Users could iterate through all the DSNPTypes. If a user passes a type to `dequeue` that isn't in the store, it will return null indicating no messages of that type.

If we wanted to check if something is in the queue without removing it, we could implement `peek()` as @enddynayn [suggested above](https://github.com/LibertyDSNP/sdk-ts/pull/40#discussion_r643464365).

src/queue/memoryQueue.ts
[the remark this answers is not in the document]
wilwade: @rosalinekarr TypeScript supports private https://www.typescriptlang.org/docs/handbook/2/classes.html#private

aneyzberg (approved):
just had one comment otherwise LGTM

shannonwells (approved):
Just the one copypasta error, approve assuming that's fixed and all other issues resolved.  The other comment is non-blocking.

src/queue/memoryQueue.ts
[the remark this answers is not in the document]
rosalinekarr: I didn't realized TypeScript supports it. Fixed now.

src/queue/memoryQueue.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  describe("#enqueue", () => {
+    it("adds a message to the queue and returns a queue id", async () => {
+      const queue = new MemoryQueue();
+
+      expect(await queue.enqueue(testMsg)).toEqual("0:0");
+    });
+  });
+
+  describe("#dequeue", () => {
+    it("adds a message to the queue and returns a queue id", async () => {
```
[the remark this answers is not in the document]
rosalinekarr: Nope, fixed.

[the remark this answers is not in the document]
rosalinekarr: Fixed.

wilwade (approved):

PR #41, opened 2021-06-01
Feature/handle availability Part 1

wilwade:
Problem
=======
Part 1
Starting the registry contract module

[link to Pivotal Tracker #178266613](https://www.pivotaltracker.com/story/show/178266613)

Solution
========
Built out the registry contract module

Change summary:
---------------
* Updated to Ethers 5.3.0
* Added a way to use the config to set the registry contract address
* contracts.registry.register
* contracts.registry.resolveHandleToId

### DIFF
--- a/src/contracts/registry.ts
+++ b/src/contracts/registry.ts
@@ -34,7 +34,7 @@ export const resolveHandleToId = async (handle: string): Promise<HexString | nul
     return (await contract.resolveHandleToId(handle)).toHexString();
   } catch (e) {
     const vmError = getVmError(e);
-    if (vmError === "Error: VM Exception while processing transaction: revert Handle does not exist") {
+    if (vmError?.includes("Handle does not exist")) {
       return null;
     }
     throw e;
--- a/src/contracts/registry.ts
+++ b/src/contracts/registry.ts
@@ -1,0 +1,51 @@
+import { getContractAddress } from "./contract";
+import { HexString } from "../types/Strings";
+import { getConfig } from "../config/config";
+import { ContractTransaction } from "ethers";
+import {MissingContract, MissingProvider, MissingSigner} from "../utilities";
+import { Registry__factory } from "../types/typechain";
+import {Logger} from "ethers/lib/utils";
+
+const CONTRACT_NAME = "Registry";
+
+/**
+ * resolveHandleToId() Try to resolve a handle into a DSNP Id
+ *
+ * @param handle String handle to resolve
+ * @returns The Hex for the DSNP Id or null if not found
+ */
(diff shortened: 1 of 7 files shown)


src/contracts/registry.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param e The error expected to have a vm Error
+ *
+ * @returns the error if any
+ */
+const getVmError = (e: { body?: string }): string | undefined => {
+  try {
+    const parsed = JSON.parse(e.body || "{}");
+    return parsed?.error?.message;
+  } catch (e) {
+    return undefined;
```
enddynayn, line 21: how can this be reached?

src/contracts/registry.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+/**
+ * register() registers a handle to get a new DSNP Id
+ *
+ * @param identityContractAddress Address of the identity contract to use
+ * @param handle The string handle to register
+ * @returns The contract Transaction
+ */
+export const register = async (identityContractAddress: HexString, handle: string): Promise<ContractTransaction> => {
```
enddynayn, line 51: question: do users have the ability to set default identityContractAddress in the config?

enddynayn (approved):

src/contracts/registry.ts
wilwade: If the error is not a VM error it will fail to parse.

src/contracts/registry.ts
wilwade: I wonder if I should change this name. This is the full identity contract or the proxy contract, not the logic contract. So this is unique to the user.

enddynayn: Ohhh ok I see.

src/contracts/registry.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const CONTRACT_NAME = "Registry";
+
+/**
+ * Get the JSON RPC error from the body, if one exists
+ * @param e The error expected to have a vm Error
+ *
+ * @returns the error if any
+ */
+const getVmError = (e: { body?: string }): string | undefined => {
```
rosalinekarr, line 16: I wonder if this should be extracted to a utility file somewhere. It seems like it could be used for other contract methods in the future.

rosalinekarr (approved):
LGTM, only one non-blocking suggestion

src/contracts/registry.ts
wilwade: I thought about that as well, but I am hoping that we can switch to Solidity custom errors soon and it will not be so nasty. https://www.pivotaltracker.com/story/show/178337443

https://blog.soliditylang.org/2021/04/21/custom-errors/

PR #43, opened 2021-06-02
Feature/handle availability part 2 178266613

wilwade:
Problem
=======
Part 2
Handles methods for checking availability of a given handle

[link to Pivotal Tracker #178266613](https://www.pivotaltracker.com/story/show/178266613)

Solution
========
availabilityFilter and isAvailable

Change summary:
---------------
* Added handles.availabilityFilter and handles.isAvailable methods
* Updated the configuration documentation


src/config/config.ts
```diff
@@ -43,11 +55,15 @@ export const getConfig = (overrides?: Partial<Config>): Config => {
...
 /**
- * setConfig() sets the current configuration with the given object. This method
- * is not yet implemented.
+ * setConfig() sets the current configuration with the given object.
  *
  * @params newConfig The configuration settings to set with
  */
-export const setConfig = (newConfig: Config): void => {
-  config = newConfig;
+export const setConfig = (newConfig: Partial<Config>): void => {
```
acruikshank, line 62: This is great!

src/contracts/announcement.test.ts
```diff
@@ -1,31 +1,11 @@
...
-  await snapshotHardhat(provider);
-});
-
-afterEach(async () => {
-  await revertHardhat(provider);
-});
+import { setupConfig } from "../test/sdkTestConfig";
 
 describe("#batch", () => {
+  beforeAll(setupConfig);
```
acruikshank, line 7: ❤️

jest.config.ts
```diff
@@ -16,6 +16,7 @@ module.exports = {
   testResultsProcessor: "jest-junit",
   transform: { "^.+\\.(ts|tsx)$": "ts-jest" },
   testMatch: ["**/*.test.(ts|js)"],
+  setupFilesAfterEnv: ["<rootDir>/src/test/chainSetup.ts"],
```
acruikshank, line 19: This is cool and super clean, but it's also magic-y. It's not going to be obvious where a test is getting its RPC_URL or what is happening when a there's a problem with snapshots. Is there a way to include chainSetup explicitly in the test files?  If not, some explanation in the readme and in the chainSetup.ts file would help.

acruikshank (approved):
This is a great refactor. The feature also looks good.

I do have some concerns that this PR hide some critical, non-standard test logic which could cause a lot of difficulties for developers that don't understand what's going on. Making an explicit call could provide a good tradeoff between clean test setup and discoverability. Documentation also might help here.

jest.config.ts
wilwade: I agree. I'll try a way to make it an explicit include, but there might be issues with order. Either way I'll make sure to document in the test about both chainSetup and configSetup.

PR #47, opened 2021-06-03
add methods to create and register identity

acruikshank:
Problem
=======
We need a way to create a user by creating a beacon proxy contract and registering it to a handle.
[link to Pivotal Tracker #178267684](https://www.pivotaltracker.com/story/show/178267684)

Solution
========
This PR adds a method to call to `createAndRegisterBeaconProxy` on the `BeaconFactory` contract, and a porcelain method that calls this function and waits for the result to land on chain.

Change summary:
---------------
1. upgrade to latest contracts
2. add contract logic to parse logs
3. add plumbing method to call createAndRegisterBeaconProxy
4. add plumbing method to call the above, wait for the transaction to land, and return the Id.

### DIFF
--- a/src/contracts/contract.ts
+++ b/src/contracts/contract.ts
@@ -1,10 +1,26 @@
 //eslint-disable-next-line
 import { HexString } from "../types/Strings";
 import { keccak256 } from "js-sha3";
+import * as types from "../types/typechain";
 import { ethers } from "ethers";
+import { JsonFragment } from "@ethersproject/abi";
 
 export const keccakTopic = (topic: string): HexString => "0x" + keccak256(topic);
 const DSNP_MIGRATION_TYPE = "DSNPMigration(address,string)";
+
+type RawLog = { topics: Array<string>; data: string };
+
+const EVENTS_ABI = new ethers.utils.Interface(
+  [
+    types.Announcer__factory,
+    types.BeaconFactory__factory,
+    types.Identity__factory,
+    types.Migrations__factory,
+    types.Registry__factory,
+  ]
+    .reduce((m, f) => m.concat(f.abi as Array<JsonFragment>), [] as Array<JsonFragment>)
+    .filter((ef) => ef.type === "event") as Array<JsonFragment>
+);
 
 interface ContractResult {
   contractAddr: string;
@@ -88,4 +104,31 @@ export const getVmError = (e: { body?: string; error?: { body?: string } }): str
     return undefined;
   }
   return undefined;
+};
+
+/**
+ * Parse all transaction logs.
+ * This requires that all contracts involved in processing the transaction be included in EVENTS_ABI.
+ * @param logs raw logs from a transaction
+ * @return parsed logs excluding any logs that cannot be parsed by the interface.
+ * @throws error if a log is unparsable. This is probably because the event's ABI has not been added to EVENTS_ABI.
+ */
+export const parseLogs = (logs: Array<RawLog>): Array<ethers.utils.LogDescription> => {
+  return logs.map((l) => EVENTS_ABI.parseLog(l)) as Array<ethers.utils.LogDescription>;
+};
+
+/**
+ * Find event with given name.
+ * @param name name of event to find.
+ * @param logs raw logs from a transaction
+ * @return First event in log that matches name
+ * @throws error if no matching events were found
+ * @throws error if a log is unparsable. This is probably because the event's ABI has not been added to EVENTS_ABI.
+ */
+export const findEvent = (name: string, logs: Array<RawLog>): ethers.utils.LogDescription => {
+  const matches = parseLogs(logs).filter((e) => e.name === name);
+  if (matches.length < 1) {
+    throw `no ${name} logs found`;
+  }
+  return matches[0];
 };
--- a/src/contracts/contract.ts
+++ b/src/contracts/contract.ts
@@ -126,9 +126,9 @@ export const parseLogs = (logs: Array<RawLog>): Array<ethers.utils.LogDescriptio
  * @throws error if a log is unparsable. This is probably because the event's ABI has not been added to EVENTS_ABI.
  */
 export const findEvent = (name: string, logs: Array<RawLog>): ethers.utils.LogDescription => {
-  const matches = parseLogs(logs).filter((e) => e.name === name);
-  if (matches.length < 1) {
+  const event = parseLogs(logs).find((e) => e.name === name);
+  if (event === undefined) {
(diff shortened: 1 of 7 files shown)


src/contracts/identity.ts
```diff
@@ -67,6 +72,25 @@ export const createBeaconProxyWithOwner = async (
...
+ * @param handle The string handle to register
+ * @returns The contract Transaction
+ */
+export const createAndRegisterBeaconProxy = async (
+  userAddress: EthereumAddress,
+  handle: string
+): Promise<ContractTransaction> => {
+  const beaconFactory = await getBeaconFactoryContract();
+  const beaconAddr = await getBeaconAddress();
+  const { signer } = getConfig();
```
enddynayn, line 89: do we want to pass config options?

enddynayn (approved):

src/contracts/contract.ts
```diff
@@ -68,3 +82,32 @@ export const getContractAddress = async (
...
+/**
+ * Find event with given name.
+ * @param name name of event to find.
+ * @param logs raw logs from a transaction
+ * @return First event in log that matches name
+ * @throws error if no matching events were found
+ * @throws error if a log is unparsable. This is probably because the event's ABI has not been added to EVENTS_ABI.
+ */
+export const findEvent = (name: string, logs: Array<RawLog>): ethers.utils.LogDescription => {
+  const matches = parseLogs(logs).filter((e) => e.name === name);
```
enddynayn, line 108: what about using .find()?

src/social/handles.ts
```diff
@@ -39,6 +42,22 @@ export const getUser = async (_id: Handle, _opts?: Config): Promise<User> => {
...
+ * createUser() creates a new identity for a public key and registers a handle to it.
+ * This function will wait for the identity to land on chain before resolving.
+ * @param addr public key address that will be used to control identity delegate
+ * @param handle name of identity (must be globaly unique)
+ * @param opts Optional. Configuration overrides, such as from address, if any
+ * @return id of identity created
+ */
+export const createUser = async (addr: HexString, handle: Handle): Promise<number> => {
+  const txn = await createAndRegisterBeaconProxy(addr, handle);
+  const receipt = await txn.wait(1);
```
enddynayn, line 55: could this be part of a config?

src/contracts/identity.ts
```diff
@@ -130,3 +154,15 @@ export const isAuthorizedTo = async (
 
   return Identity__factory.connect(contractAddress, provider).isAuthorizedTo(address, permission, blockNumber);
 };
+
+const getBeaconAddress = async () => {
+  const {
+    provider,
+    contracts: { beacon },
+  } = getConfig();
```
enddynayn, line 162: config option?

src/contracts/identity.ts
```diff
@@ -130,3 +154,15 @@ export const isAuthorizedTo = async (
 
   return Identity__factory.connect(contractAddress, provider).isAuthorizedTo(address, permission, blockNumber);
 };
+
+const getBeaconAddress = async () => {
```
enddynayn, line 158: question for me: when do we add the return type in typescript

enddynayn:
fabuloso.

src/contracts/identity.test.ts
```diff
@@ -125,4 +128,36 @@ describe("identity", () => {
...
+  describe("#createAndRegisterBeaconProxy", () => {
+    const handle = "flarp";
+    const fakeAddress = "0x1Ea32de10D5a18e55DEBAf379B26Cc0c6952B168";
+
+    it("returns a Contract Transaction that can be resolved into a DSNP Id", async () => {
+      const transaction = await createAndRegisterBeaconProxy(fakeAddress, handle);
+
+      const receipt = await transaction.wait(1);
+
+      expect(getIdFromReceipt(receipt).toNumber()).toBeGreaterThan(999);
```
aneyzberg, line 141: Just curious - why is this expectation greater than 999?

src/contracts/identity.ts
```diff
@@ -67,6 +72,25 @@ export const createBeaconProxyWithOwner = async (
...
+ */
+export const createAndRegisterBeaconProxy = async (
+  userAddress: EthereumAddress,
+  handle: string
+): Promise<ContractTransaction> => {
+  const beaconFactory = await getBeaconFactoryContract();
+  const beaconAddr = await getBeaconAddress();
+  const { signer } = getConfig();
+  if (!signer) throw MissingSigner;
+  return await beaconFactory.connect(signer).createAndRegisterBeaconProxy(beaconAddr, userAddress, handle);
```
aneyzberg, line 91: Do we need the `.connect(signer)` here? I think `getBeaconFactoryContract` returns a contract instance that should already be connected to?.

src/contracts/identity.ts
aneyzberg: @enddynayn I believe it should have a return type.

aneyzberg (approved):
just a couple small comments - otherwise lgtm

src/contracts/contract.ts
acruikshank: good call. Fixed.

src/contracts/identity.ts
acruikshank: This follows the pattern of other contract retrieval functions in `identity.ts` that expect the config to already have been set. I'd rather not expose it here because I'd then have the same issue for `createAndRegisterBeaconProxy`, and so on.

src/contracts/identity.ts
acruikshank: Thank you. Added.

src/contracts/identity.test.ts
acruikshank: The registry contract starts ids at 1000. So we know it shouldn't be less, but we don't know much more than that. I added a comment to this effect.

src/contracts/identity.ts
acruikshank: We do not! Nice catch.

src/social/handles.ts
acruikshank: This is a good point. There aren't going to be forks on the initial PoA network, so 1 is always sufficient. For that reason, I'd like to push off a configuration parameter until we have a better idea how config is going to be used.

src/contracts/identity.ts
acruikshank: This line is no longer needed. Configuration options are used by some of the methods lower down. We're currently not passing in config for other methods in identity, so I'm going to stick to the pattern. We may decide to change all of them later.

PR #48, opened 2021-06-04
ls-remote does not always get all the commits

wilwade:
When squash-merging just a single commit, it is different than multiple commits. Thanks GitHub.


aneyzberg (approved):
lgtm

PR #54, opened 2021-06-07
Features/activity pub improvements 178266808

rosalinekarr:
Problem
=======
We need some tweaks to the activity pub helpers for the announcement porcelain story, but including it there would result in a huge git diff for review.
[#178266808](https://www.pivotaltracker.com/story/show/178266808)

Solution
========
Pull the activity pub changes out into a separate PR for easier review

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,7 +18,7 @@
   },
   "rules": {
     "semi": ["error", "always"],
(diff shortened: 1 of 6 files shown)


aneyzberg (approved):
lgtm

PR #56, opened 2021-06-08
Validate messages, signature + permission #178267391

shannonwells:
Problem
=======
>As the App API, I need to know if a public key has a given permission on a given contract So that I can test a generic signature against a DSNP Id

[Pivotal Tracker #178267391](https://www.pivotaltracker.com/story/show/178267391)

Solution
========
* new plumbing function contracts/registry.ts#validateMessage
* tests for same
* added some test helpers in testAccounts.ts which let you set up a new signer and easily create a new DSNP identity with a new handle all in one go, using the specified test account by its index.

Steps to Verify:
----------------
1. tests should all pass

### DIFF
--- a/src/core/contracts/registry.test.ts
+++ b/src/core/contracts/registry.test.ts
@@ -221,6 +221,13 @@ describe("registry", () => {
       expect(regs[1].handle).toEqual(handle + "new");
     });
   });
+
+  describe("validateMessage", () => {
+    it("throws if not registered");
+    it("throws if can't get block height");
+    it("returns true if the signer is authorized to do the thing");
+    it("returns false if the signer is not authorized to do the thing");
+  });
 });
 
 const getIdFromRegisterTransaction = async (transaction: ContractTransaction) => {
--- a/src/core/contracts/registry.test.ts
+++ b/src/core/contracts/registry.test.ts
@@ -14,6 +14,11 @@ import { setupConfig, getSignerFromPrivateKey } from "../../test/sdkTestConfig";
 import { Permission } from "./identity";
 import { DSNPMessage, sign } from "../messages/messages";
 import { generateBroadcast} from "../../test/generators/dsnpGenerators";
+import {
+  getIdFromRegisterTransaction,
+  newRegistrationForAccountIndex,
+  RegistrationWithSigner,
+} from "../../test/testAccounts";
 
 describe("registry", () => {
   let signer: Signer;
@@ -235,9 +240,11 @@ describe("registry", () => {
   describe("validateMessage", () => {
     const msg: DSNPMessage = generateBroadcast();
 
-    let contractAddr = "";
     const permAllowed = Permission.ANNOUNCE;
     const permDenied = Permission.OWNERSHIP_TRANSFER;
+    let contractAddr = "";
+    let sig = "";
+    let dsnpId = "";
     let signerAddr = "";
 
     beforeAll(async () => {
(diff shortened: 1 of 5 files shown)


src/core/contracts/registry.test.ts
```diff
@@ -221,11 +237,68 @@ describe("registry", () => {
       expect(regs[1].handle).toEqual(handle + "new");
     });
   });
-});
 
-const getIdFromRegisterTransaction = async (transaction: ContractTransaction) => {
```
shannonwells, line 226: this was moved to the testAccounts file.

src/test/testAccounts.ts
```diff
@@ -0,0 +1,159 @@
...
+  if (!provider) throw new Error("no provider configured");
+  return new ethers.Wallet(TestAccounts[accountIndex].privateKey, provider);
+};
+
+/**
+ * Parses a DSNP Id in a contract transaction log event.
+ * @param transaction
+ * @returns the DSNP Id
+ */
+export const getIdFromRegisterTransaction = async (transaction: ContractTransaction): Promise<any> => {
```
rosalinekarr, line 27: `Promise<any>` => `Promise<DSNPId>`?

src/core/contracts/registry.ts
```diff
@@ -102,6 +107,36 @@ export const getDSNPRegistryUpdateEvents = async (
...
+): Promise<boolean> => {
+  const reg = await resolveId(dsnpId);
+  if (!reg) throw MissingContract;
+
+  const { provider } = getConfig();
+  if (!provider) throw MissingProvider;
+
+  const realBlockNumber = blockNumber ? blockNumber : "latest";
+  const bh = (await provider?.getBlock(realBlockNumber))?.number;
+  if (!bh) throw new Error("could not get block at height " + realBlockNumber);
```
wilwade, line 134: Do we need to do this test?

isAuthorizedTo can take future block numbers and the "forever" version is with block="0x0" which would be a fine default.

src/test/testAccounts.ts
rosalinekarr: I added a `DSNPId` type in [my PR](https://github.com/LibertyDSNP/sdk-ts/pull/42/files#diff-4a0c39af59da784a31b673fe4bb67584304827189fc08f23feefd17bdaf4239eR6). Feel free to copy paste it and I'll just rebase when this PR gets merged in.

src/test/testAccounts.ts
```diff
@@ -0,0 +1,159 @@
...
+  const newSignerAddr = await signer.getAddress();
+  const identityContract = await new Identity__factory(signer).deploy(newSignerAddr);
+  await identityContract.deployed();
+  const contractAddr = identityContract.address;
+  const tx2 = await register(contractAddr, handle);
+  const dsnpId = await getIdFromRegisterTransaction(tx2);
+  return { contractAddr, dsnpId, handle, signer };
+};
+
+const TestAccounts = [
```
rosalinekarr, line 55: Not blocking, but maybe we should move this to the top of the files and all-caps the variable name since it seems like a constant.

src/core/contracts/registry.ts
```diff
@@ -102,6 +107,36 @@ export const getDSNPRegistryUpdateEvents = async (
...
+ * specified.
+ * @param signature the signature for the message
+ * @param message the DSNPMessage
+ * @param dsnpId the DSNP Id of the supposed signer
+ * @param permission the permissions to check for
+ * @param blockNumber (optional) the block number at which to perform this check.  Defaults to "latest"
+ */
+export const validateMessage = async (
+  signature: HexString,
+  message: DSNPMessage,
```
wilwade, line 121: I think we need a non-DSNPMessage version of this method as well. Could we support it like this:

```suggestion
  message: DSNPMessage | string,
```
...
```typescript
const serialized = typeof message === string ? message : serialize(message);
```

rosalinekarr (approved):
One blocking item, one non-blocking, otherwise LGTM

wilwade (changes requested):
One check I think we can remove, and an opening up of a type.

src/core/contracts/registry.ts
shannonwells: This was a deliberate choice; I'd prefer to make a separate generic string message validation because of the sorting.

src/core/contracts/registry.ts
shannonwells: Ah ok I didn't know that.

wilwade (approved):
🥇

src/test/testAccounts.ts
shannonwells: Unfortunately it's not always a string.  Sometimes it's a BigNumber.  Changing it to one or the other fails tests.  This function was already written and I moved it, then added a return type to address the existing warning... which of course created a different warning.  But fixing it properly is a little more than what I felt like doing in this PR.

PR #58, opened 2021-06-09
Pre release QA Part 2

wilwade:
Problem
=======
Could not access core/messages

Solution
========
Export core/messages

Change summary:
---------------
* Rename Storage -> Store
* export core/message


.gitignore
```diff
@@ -15,7 +15,7 @@ dist
 *.js
 *.d.ts
 docs
-core
+/core
```
shannonwells, line 18: 😠  stupid that you had to do this.

src/core/index.ts
```diff
@@ -18,11 +18,14 @@ export const batch = batchImport;
...
 import * as queueImport from "./queue";
 export const queue = queueImport;
 
-import * as storageImport from "./storage";
-export const storage = storageImport;
+import * as storeImport from "./store";
+export const store = storeImport;
 
 import * as utilitiesImport from "./utilities";
 export const utilities = utilitiesImport;
```
shannonwells, line 31: is there an 'import * as the filename I imported it + Import and then export it as itself' ? probably not, this is TS/JS!

shannonwells (approved):

src/core/index.ts
```diff
@@ -18,11 +18,14 @@ export const batch = batchImport;
 import * as contractsImport from "./contracts";
 export const contracts = contractsImport;
 
+import * as messagesImport from "./messages";
+export const messages = messagesImport;
```
wilwade, line 22: This is the one other piece. Adding this to the exports out of the package

enddynayn (commented):
👍

.gitignore
wilwade: Agreed. I think we are going to have to build the package with a separate package.json in a different directory. @aneyzberg and I looked at it, but we backed out. It is not super quick and causes other issues that take time to resolve (like rebuilding the package.json with the correct stuff and in the correct paths)

PR #61, opened 2021-06-10
getters & refactors

shannonwells:
Problem
=======
Add getter methods to config module [link to Pivotal Tracker #178458945](https://www.pivotaltracker.com/story/show/178458945)

Solution
========
Proposed solution:
1. A bunch of `requireFoo` functions which pretty much replace all getters that are being used.  They always throw if Foo is undefined.
1. A requireGetConfig function to use if you want to guarantee multiple things are there.  Pass a list of strings (feels yucky, alternatives requested) and your configOpts.
1. Tests for above

I did not completely do a refactor because I wanted input before I did the rest.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,6 +1,15 @@
 import { providers, Wallet } from "ethers";
-import { getConfig, setConfig, Config, getQueue, ConfigOpts, requireGetConfig } from "./config";
-import { MissingStore, MissingUser } from "./core/utilities";
+import {
+  getConfig,
+  setConfig,
+  Config,
+  getQueue,
+  ConfigOpts,
+  MissingStore,
+  MissingUser,
+  requireGetConfig,
+} from "./config";
+import {} from "./core/utilities";
 
 describe("config", () => {
   describe("#getConfig", () => {
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -1,13 +1,15 @@
-import { getConfig, setConfig, Config } from "./config";
+import { providers, Wallet } from "ethers";
+import { getConfig, setConfig, Config, getQueue, ConfigOpts, requireGetConfig } from "./config";
+import { MissingStore, MissingUser } from "./core/utilities";
 
 describe("config", () => {
   describe("#getConfig", () => {
     it("returns the default settings when the config hasn't been set yet", () => {
-      const results = getConfig();
+      const queue = getQueue();
 
       // We have to test keys here because #toMatchObject doesn't work with
       // object instances
-      expect(results["queue"]).toBeInstanceOf(Object);
+      expect(queue).toBeInstanceOf(Object);
     });
 
     it("fetches the current config settings", () => {
@@ -47,6 +49,47 @@ describe("config", () => {
       setConfig(testConfig);
 
       expect(getConfig()).toMatchObject({ test: "object" });
+    });
+  });
+
+  describe("getConfigOrThrow", () => {
+    // get a minimal valid config
+    describe("when config is missing required params", () => {
+      it("throws when something is missing", () => {
+        const testConfig = {
+          signer: Wallet.createRandom(),
+          provider: new providers.JsonRpcProvider("http://localhost:8383"),
+        };
+        setConfig(testConfig);
+        expect(() => requireGetConfig(["signer", "provider", "currentFromId"])).toThrow(MissingUser);
+      });
+
+      it("throws when we require invalid config", () => {
+        expect(() => requireGetConfig(["singer"])).toThrow("unknown config: singer");
+      });
+
+      it("throws when the config key exists but its value is undefined", () => {
+        setConfig({
+          provider: undefined,
+          store: undefined,
+          signer: undefined,
+          currentFromId: undefined,
+        });
+        expect(() => requireGetConfig(["store", "currentFromId", "provider", "signer"])).toThrow(MissingStore);
+      });
+    });
+
+    describe("when config has all expected values", () => {
+      const testConfig: ConfigOpts = getConfig();
+
+      [
+        { name: "when passed no params", require: [] },
+        { name: "when passed params that exist", require: ["queue", "contracts"] },
+      ].forEach((tc) => {
+        it(`${tc.name} returns config`, () => {
+          expect(requireGetConfig([])).toEqual(testConfig);
+        });
+      });
     });
   });
 });
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,6 +4,22 @@ import MemoryQueue from "./core/queue/memoryQueue";
 import { QueueInterface } from "./core/queue";
 import { StoreInterface } from "./core/store";
 import { HexString } from "./types/Strings";
+import { MissingProvider, MissingSigner, MissingStore, MissingUser } from "./core/utilities";
+
+export interface Contracts {
+  /** The Address of the Batch Announce contract */
+  announcer?: HexString;
+  /** The Address of the Beacon contract */
+  beacon?: HexString;
+  /** The Address of the Beacon Proxy Factory contract */
+  beaconFactory?: HexString;
+  /** The Address of the Identity Logic contract */
+  identityLogic?: HexString;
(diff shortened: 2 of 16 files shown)


src/config.ts
```diff
@@ -4,6 +4,22 @@ import MemoryQueue from "./core/queue/memoryQueue";
 import { QueueInterface } from "./core/queue";
 import { StoreInterface } from "./core/store";
 import { HexString } from "./types/Strings";
+import { MissingProvider, MissingSigner, MissingStore, MissingUser } from "./core/utilities";
```
rosalinekarr, line 7: Maybe we should just pull these errors into this file now that this is the only place they're thrown?

src/config.ts
```diff
@@ -77,3 +82,63 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
     ...newConfig,
   });
 };
+
+// - Update all `getConfig()` calls to use these getters where appropriate
+// - Move associated "Missing" errors to the config module
+export const requireGetProvider = (opts?: ConfigOpts): ethers.providers.Provider => {
```
rosalinekarr, line 88: Can you throw some typedoc comments on these methods?

src/config.ts
```diff
@@ -4,6 +4,22 @@ import MemoryQueue from "./core/queue/memoryQueue";
...
+  beacon?: HexString;
+  /** The Address of the Beacon Proxy Factory contract */
+  beaconFactory?: HexString;
+  /** The Address of the Identity Logic contract */
+  identityLogic?: HexString;
+  /** The Address of the Identity Clone Proxy Factory contract */
+  identityCloneFactory?: HexString;
+  /** The Address of the Registry contract */
+  registry?: HexString;
+}
```
rosalinekarr, line 22: ❤️

rosalinekarr (changes requested):
Two things, otherwise looks good

src/config.ts
shannonwells: Yes, I meant actually to make this a draft PR and forgot.

enddynayn (approved):

enddynayn:
❤️ superb!

src/config.ts
rosalinekarr: ah, oops

src/config.ts
```diff
@@ -77,3 +87,89 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
...
+  const c = getConfig(opts);
+  return c.contracts;
+};
+
+/**
+ * Get the full configuration. Use if you need to guarantee that >1 configs are set
+ * @param requiredConfigs: list of config strings
+ * @param opts
+ */
+export const requireGetConfig = (requiredConfigs?: Array<string>, opts?: ConfigOpts): Config => {
```
enddynayn, line 154:
```
const config = getConfig(opts);

const requiredConfigs = {
  store: MissingStore,
  signer: MissingSigner,
  provider: MissingProvider,
  currentFromId: MissingUser,
}

requiredConfigs
  .filter((configKey) => !config[configKey])
  .forEach( configKey => {
    if (requiredConfigs[configKey]) throw requiredConfigs[configKey];
    
    throw new Error("unknown config: " + configKey);
})

return config;

```

This might be a little easier on the eyeballs if it works. 🤷

src/config.ts
```diff
@@ -77,3 +87,89 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
...
+  const c = getConfig(opts);
```
enddynayn, line 155: This comment is the only one that means something to me. When I was reading the code above which contain two lines, I overlooked the short abbreviations. But here when I am reading through the code that has more involved logic this short abbreviation kills the mental flow and readability for me.

I know that this is something small but can we please not use short abbreviations, at least, when there are more than two lines of code?

❤️ 🙏 por favor 🙏 ❤️

src/config.ts
```diff
@@ -77,3 +87,89 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
     ...newConfig,
   });
 };
+
+/**
+ * Get the provider and if undefined, throw.
+ * @param opts overrides for the current configuration.
```
rosalinekarr, line 93: Can we document the `@returns` too?

src/config.ts
```diff
@@ -77,3 +87,89 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
...
+ */
+export const requireGetProvider = (opts?: ConfigOpts): ethers.providers.Provider => {
+  const c = getConfig(opts);
+  if (!c.provider) throw MissingProvider;
+  return c.provider;
+};
+
+/**
+ * Get the signer and if undefined, throw.
+ * @param opts overrides for the current configuration.
```
rosalinekarr, line 103: Same here and all the following typedoc comments

src/config.ts
```diff
@@ -77,3 +87,89 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
...
+  const c = getConfig(opts);
+  if (!c.store) throw MissingStore;
+  return c.store;
+};
+
+/**
+ * Get the currentFromId and if undefined, throw.
+ * @param opts overrides for the current configuration.
+ */
+export const requireGetCurrentFromId = (opts?: ConfigOpts): string => {
```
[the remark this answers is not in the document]
rosalinekarr: I like the distinction between getters that do and don't throw. Maybe it would better if the names only had either a `get` prefix or a `require` prefix? Like, this one could be `requireCurrentFromId` instead.

wilwade (commented):
I worry a bit about requireGetConfig, but so much better error handling!

rosalinekarr (approved):
Only non-blocking feedback, otherwise LGTM

src/core/contracts/identity.ts
```diff
@@ -96,9 +96,8 @@ const getIdentityLogicContractAddress = async (opts?: ConfigOpts): Promise<Ether
   const {
     provider,
     contracts: { identityLogic },
-  } = getConfig(opts);
-  if (!provider) throw MissingProvider;
-  const address = identityLogic || (await getContractAddress(provider, IDENTITY_CONTRACT));
+  } = requireGetConfig(["provider"], opts);
```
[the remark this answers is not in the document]
shannonwells: good catch, I missed one.

src/config.ts
shannonwells: sure thing 😸

src/config.test.ts
```diff
@@ -1,13 +1,24 @@
...
+  getConfig,
+  setConfig,
+  Config,
+  getQueue,
+  ConfigOpts,
+  MissingStore,
+  MissingUser,
+  requireGetConfig,
+} from "./config";
+import {} from "./core/utilities";
```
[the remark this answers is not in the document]
shannonwells: I just forgot to delete the line ....

src/config.ts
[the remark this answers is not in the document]
shannonwells: I didn't add plain getters for everything, only because they weren't needed anywhere. I did not make everything 'require' because for queue and contracts, those are not optional fields and would never throw.

 This is not meant to be a final version.  The intent was to add non-throwing getters as needed. So if it's fine to have something undefined, then there are two options; use getConfig, or create a new getter.

src/core/contracts/identity.ts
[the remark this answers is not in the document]
shannonwells: It won't work here because it only returns a provider, but this function needs the entire config.

src/core/contracts/registry.ts
```diff
@@ -159,10 +162,9 @@ const getContract = async (opts?: ConfigOpts) => {
   const {
     provider,
     contracts: { registry },
-  } = getConfig(opts);
-  if (!provider) throw MissingProvider;
-  const address = registry || (await getContractAddress(provider, CONTRACT_NAME));
+  } = requireGetConfig(["provider"], opts);
```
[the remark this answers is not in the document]
shannonwells: Same as above.

src/config.ts
shannonwells: To me this is more confusing.  Sometimes a long form is better than a short and clever form.

PR #62, opened 2021-06-11
feat(batch): add ability to write batch

enddynayn:
Add the ability to write a batch file to storage using streaming.
Writing the batch includes writing the corresponding bloom filter
for every DSNP message type.

### DIFF
--- a/src/core/batch/batch.test.ts
+++ b/src/core/batch/batch.test.ts
@@ -1,5 +1,11 @@
-import { batchIncludes } from "./batch";
-import { ParquetReader } from "@dsnp/parquetjs";
+import * as batch from "./batch";
+import { generateBroadcast } from "../../generators/dsnpGenerators";
+import { EmptyArrayError } from "../utilities/errors";
+import { putStream } from "../store/interface";
+import { BroadcastSchema } from "./parquetSchema";
+import { ParquetReader, ParquetWriter } from "@dsnp/parquetjs";
+
+jest.mock("../store/interface");
 
 describe("batchIncludes", () => {
   let reader: typeof ParquetReader;
@@ -23,8 +29,72 @@ describe("batchIncludes", () => {
   });
 
   it("calls getBloomFilter", () => {
-    batchIncludes(reader, "name", "banana");
+    batch.batchIncludes(reader, "name", "banana");
 
     expect(reader.getBloomFilters.mock.calls.length).toBe(1);
+  });
+});
+
+describe("#writeBatch", () => {
+  const messages = [generateBroadcast()];
+  const writeStream = { write: jest.fn(), end: jest.fn() };
+  const parquetWriterInstance = {
+    appendRow: jest.fn().mockImplementation(async (message) => message),
+    close: jest.fn(),
+  };
+
+  const { writeBatch } = batch;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    ParquetWriter.openStream = jest.fn().mockResolvedValue(parquetWriterInstance);
+  });
+
+  it("calls ParquetWriter#openStream to start a writable stream", async () => {
+    await writeBatch(writeStream, BroadcastSchema, messages);
+    expect(ParquetWriter.openStream).toHaveBeenCalled();
+  });
+
+  it("calls ParquetWriter#appendRow to add a row to parquet stream", async () => {
+    await writeBatch(writeStream, BroadcastSchema, messages);
+    expect(parquetWriterInstance.appendRow).toHaveBeenCalledTimes(1);
+  });
+
+  it("calls ParquetWriter#closes to end the stream", async () => {
+    await writeBatch(writeStream, BroadcastSchema, messages);
+    expect(parquetWriterInstance.close).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe("#createFile", () => {
+  const { createFile } = batch;
+  const messages = [generateBroadcast()];
+
+  beforeEach(() => {
+    jest.spyOn(batch, "writeBatch");
+    (putStream as jest.Mock).mockImplementation((path, callback) => callback(path, callback));
+  });
+
+  it("calls putStream to start streaming", async () => {
+    await createFile("batch.parquet", messages);
+
+    expect(putStream).toHaveBeenCalled();
+  });
+
+  it("calls #writeBatch to stream write parquet", async () => {
+    await createFile("batch.parquet", messages);
+
+    expect(batch.writeBatch).toHaveBeenCalledWith(
+      "batch.parquet",
+      expect.objectContaining({ fieldList: expect.any(Array) }),
+      messages,
+      { bloomFilters: { bloomFilters: [{ column: "fromId" }] } }
+    );
+  });
+
+  describe("when messages argument is empty", () => {
+    it("throws an EmptyArrayError", async () => {
+      await expect(batch.createFile("batch.parquet", [])).rejects.toThrow(EmptyArrayError);
+    });
   });
 });
--- a/src/core/batch/batch.ts
+++ b/src/core/batch/batch.ts
@@ -1,7 +1,10 @@
-import { DSNPType } from "../messages/messages";
-import { NotImplementedError } from "../utilities";
+import { DSNPMessage, DSNPType } from "../messages/messages";
 import request from "request";
-import { ParquetReader } from "@dsnp/parquetjs";
+import parquet from "@dsnp/parquetjs";
+const { ParquetReader, ParquetWriter, ParquetSchema } = parquet;
+import { putStream, ReadWriteStream, WriteStream } from "../store/interface";
+import { getSchemaFor, getBloomFilterOptionsFor, Schema, BloomFilterOptions } from "./parquetSchema";
+import { EmptyArrayError } from "../utilities";
 
 type ReadRowFunction = {
   (row: DSNPType): void;
@@ -17,17 +20,53 @@ interface BloomFilterData {
   RowGroupIndex: number;
 }
 
+interface WriteBatchFileOptions {
+  bloomFilters?: BloomFilterOptions;
+}
+
 export type BatchFileObject = string;
 
 /**
- * createFile() takes a series of DSNP messages and returns a Batch file
- * object for publishing. This method is not yet implemented.
+ * createFile() takes a series of DSNP messages and returns a URL
+ * for storage location.
  *
- * @param events  An array of DSNPEvents to include in the batch file
- * @returns         A batch file object
+ * @param targetPath  The path to and name of file
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns         A URL of the storage location
+ * @throws error if messages argument is empty.
  */
-export const createFile = (_events: DSNPType[]): BatchFileObject => {
-  throw NotImplementedError;
+export const createFile = async (targetPath: string, messages: DSNPMessage[]): Promise<URL> => {
+  if (messages.length === 0) throw EmptyArrayError;
+
+  const schema = new ParquetSchema(getSchemaFor(messages[0].dsnpType));
+  const bloomFilterOptions = getBloomFilterOptionsFor(messages[0].dsnpType);
+
+  return putStream(targetPath, async (writeStream: ReadWriteStream) => {
+    await writeBatch(writeStream, schema, messages, { bloomFilters: bloomFilterOptions });
+  });
+};
+
+/**
+ * writeBatch() takes a series of DSNP messages and returns a Batch file
+ * object for publishing.
+ *
+ * @param writeStream A writable stream
+ * @param schema    The ParquetJS schema for the messages DSNP type
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @param opts     Options for creating a Parquet file
+ * @returns        A void promise which will either resolve or reject
+ */
+export const writeBatch = async (
+  writeStream: WriteStream,
+  schema: Schema,
+  messages: DSNPMessage[],
+  opts?: WriteBatchFileOptions
+): Promise<void> => {
+  const writer = await ParquetWriter.openStream(schema, writeStream, opts);
+
+  await Promise.all(messages.map((message) => writer.appendRow(message)));
+
+  await writer.close();
 };
 
 /**
--- a/src/core/batch/parquetSchema.ts
+++ b/src/core/batch/parquetSchema.ts
@@ -3,12 +3,21 @@
  * See DSNP type definitions in DSNP.d.ts for additional documentation of fields.
  */
 
+import { DSNPType } from "../messages/messages";
+
+/**
+ * BloomFilterColumnOptions: bloom filter options for a column intented to be used with when writing a batch file
+ */
 export interface BloomFilterColumnOptions {
   column: string;
   numFilterBytes?: number;
   falsePositiveRate?: number;
   numDistinct?: number;
 }
+
+/**
+ * BloomFilterOptions: bloom filter options intented to be used with when writing a batch file
+ */
 export interface BloomFilterOptions {
   bloomFilters: Array<BloomFilterColumnOptions>;
 }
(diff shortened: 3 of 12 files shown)


src/core/batch/batch.ts
```diff
@@ -23,13 +24,34 @@ export type BatchFileObject = string;
...
+    await writeBatch(writeStream, schema, messages);
+  });
+};
+
+/**
+ * writeBatch() takes a series of DSNP messages and returns a Batch file
+ * object for publishing.
+ *
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns        A function with
```
enddynayn, line 47: docds

src/core/batch/parquetSchema.ts
```diff
@@ -83,3 +85,9 @@ export const ReactionSchema = {
 export const ReactionBloomFilterOptions = {
   bloomFilters: [{ column: "emoji" }, { column: "fromId" }, { column: "inReplyTo" }],
 };
+
+export const getSchemaFor = (dsnpType: DSNPType): Schema => {
```
enddynayn, line 89: docs

src/core/storage/s3Node.ts
```diff
@@ -1,6 +1,19 @@
 import { S3Client, S3ClientConfig, GetObjectCommand, PutObjectCommand } from "@aws-sdk/client-s3";
+import { Upload } from "@aws-sdk/lib-storage";
 import { StorageInterface } from "./storage";
-import { Readable } from "stream";
+import { Readable, PassThrough } from "stream";
+
+interface PassThroughStream {
```
enddynayn, line 6: docs

src/core/storage/s3Node.ts
```diff
@@ -1,6 +1,19 @@
...
+  read(this: unknown, size: number): void;
+  write(chunk: unknown, encoding?: string, callbck?: (error: Error | null | undefined) => void): boolean;
+  write(chunk: unknown, cb?: (error: Error | null | undefined) => void): boolean;
+  end(chunk: unknown, cb?: () => void): void;
+  end(chunk: unknown, encoding?: string, cb?: () => void): void;
+}
+
+interface PassThroughCallback {
```
enddynayn, line 14: docs

src/core/batch/batch.test.ts
```diff
@@ -26,8 +30,65 @@ describe("batchIncludes", () => {
...
+
+    expect(putStream).toHaveBeenCalled();
+  });
+
+  it("calls #writeBatch to stream write parquet", async () => {
+    await createFile("batch.parquet", messages);
+
+    expect(batch.writeBatch).toHaveBeenCalledWith(
+      "batch.parquet",
+      expect.objectContaining({ fieldList: expect.any(Array), taco: "0ne" }),
```
enddynayn, line 90: wrong

src/config.ts
```diff
@@ -1,7 +1,7 @@
+import { QueueInterface } from "core/queue/queue";
 import { ethers } from "ethers";
 
 import MemoryQueue from "./core/queue/memoryQueue";
-import { QueueInterface } from "./core/queue";
```
rosalinekarr, line 4: why did this change? i kind of prefer having the external imports separated from the internal ones.

src/core/batch/batch.ts
```diff
@@ -21,13 +24,34 @@ export type BatchFileObject = string;
...
+ * object for publishing.
  *
- * @param events  An array of DSNPEvents to include in the batch file
- * @returns         A batch file object
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns         A URL
  */
-export const createFile = (_events: DSNPType[]): BatchFileObject => {
-  throw NotImplementedError;
+export const createFile = async (targetPath: string, messages: DSNPMessage[]): Promise<URL | null> => {
```
rosalinekarr, line 32: is there a strong reason to return null here? could we check for null and throw instead?

src/core/batch/batch.ts
```diff
@@ -21,13 +24,34 @@ export type BatchFileObject = string;
...
+};
+
+/**
+ * writeBatch() takes a series of DSNP messages and returns a Batch file
+ * object for publishing.
+ *
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns        A function with
+ */
+export const writeBatch = async (writeStream: WriteStream, schema: Schema, messages: DSNPMessage[]): Promise<void> => {
```
enddynayn, line 49: pass optional args

src/core/batch/batch.ts
```diff
@@ -21,13 +24,34 @@ export type BatchFileObject = string;
...
+  const writer = await ParquetWriter.openStream(schema, writeStream);
```
enddynayn, line 50: pass optional args into here

src/core/batch/batch.ts
rosalinekarr, line 49: do we have to export this method? what would be the use case for it?

rosalinekarr (commented):
looks good so far, just a few minor comments and questions

src/core/batch/batch.ts
enddynayn: I was not sure if we should. I will update it to throw.

src/core/batch/batch.ts
enddynayn: We don't have to. It made it easy to unit test?

src/config.ts
enddynayn: I have no idea. My editor was complaining and I just had it do its thang. Let me take a look.

src/core/batch/batch.ts
```diff
@@ -21,13 +24,34 @@ export type BatchFileObject = string;
...
+    await writeBatch(writeStream, schema, messages);
+  });
+};
+
+/**
+ * writeBatch() takes a series of DSNP messages and returns a Batch file
+ * object for publishing.
+ *
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns        A function with
```
enddynayn, line 47: what am i even saying here

src/config.ts
enddynayn: It looks like my editor needed to be refresh. thanks for catching this.

src/core/batch/batch.ts
```diff
@@ -17,17 +21,49 @@ interface BloomFilterData {
...
 
 /**
  * createFile() takes a series of DSNP messages and returns a Batch file
- * object for publishing. This method is not yet implemented.
+ * object for publishing.
  *
- * @param events  An array of DSNPEvents to include in the batch file
- * @returns         A batch file object
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns         A URL
```
shannonwells, line 35: this could be clearer, because it says it returns a Batch file object above when it returns a URL to the file, right?

src/core/batch/parquetSchema.ts
```diff
@@ -69,6 +84,13 @@ export const ProfileSchema = {
   signature: { type: "BYTE_ARRAY" },
 };
 
+/**
+ * ProfileBloomFilter: bloom filter options for batching profile messages
+ */
+export const ProfileBloomFilterOptions = {
```
shannonwells, line 97: 👍

src/core/batch/parquetSchema.ts
```diff
@@ -80,6 +102,39 @@ export const ReactionSchema = {
...
+/**
+ * getSchemaFor() takes DSNPType and returns its corresponding parquet schema
+ *
+ * @param   dsnpType a dsnpType
+ * @returns The corresponding parquet schema
+ */
+export const getSchemaFor = (dsnpType: DSNPType): Schema => {
+  const schemas = [GraphChangeSchema, BroadcastSchema, ReplySchema, ReactionSchema, ProfileSchema];
+
+  return schemas[dsnpType - 1];
```
shannonwells, line 128: pretty slick!  the only thing I worry about is if the order of these in dsnp types ever changes.  hopefully the tests cover this, but I'd feel a little better if it were a key/value lookup.

src/core/store/s3Node.test.ts
```diff
@@ -10,6 +15,52 @@ describe("S3Node", () => {
...
+    beforeEach(() => {
+      callback = jest.fn().mockImplementation();
+      Upload.prototype.done = jest.fn();
+    });
+
+    it("calls callback with PassThrough stream", async () => {
+      await client.putStream("batch.parquet", callback);
+
+      expect(callback).toHaveBeenCalledWith(
+        expect.objectContaining({
```
shannonwells, line 30: 🥇

shannonwells (approved):
I have just the one concern but I'll leave the decision up to you.

src/core/batch/batch.test.ts
```diff
@@ -1,5 +1,13 @@
-import { batchIncludes } from "./batch";
-import { ParquetReader } from "@dsnp/parquetjs";
+import * as batch from "./batch";
+import { generateBroadcast } from "../../generators/dsnpGenerators";
+import { EmptyArrayError } from "../utilities/errors";
+import { putStream } from "../store/interface";
+import { BroadcastSchema } from "./parquetSchema";
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const parquet = require("@dsnp/parquetjs");
+const { ParquetReader, ParquetWriter } = parquet;
```
wilwade, line 8: This should work now
```suggestion
import { ParquetReader, ParquetWriter } from "@dsnp/parquetjs";
```

src/core/batch/batch.ts
```diff
@@ -1,7 +1,11 @@
-import { DSNPType } from "../messages/messages";
-import { NotImplementedError } from "../utilities";
+import { DSNPMessage, DSNPType } from "../messages/messages";
 import request from "request";
-import { ParquetReader } from "@dsnp/parquetjs";
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const parquet = require("@dsnp/parquetjs");
+const { ParquetReader, ParquetWriter, ParquetSchema } = parquet;
```
wilwade, line 5: Should also be able to use the import version.

src/core/batch/parquetSchema.ts
```diff
@@ -3,12 +3,21 @@
  * See DSNP type definitions in DSNP.d.ts for additional documentation of fields.
  */
 
+import { DSNPType } from "../messages/messages";
+
+/**
+ * BloomFilterColumnOptions: bloom filter options for a column intented to be used with when writing a batch file
+ */
```
wilwade, line 10: ❤️ Thanks for all these docs.

src/core/batch/parquetSchema.ts
wilwade: An interesting alternative would actually be to gain type safety and would throw an error when a new DSNPType was added (make sure not to use a `case default` and TypeScript will test for completeness on enums):

```suggestion  
export const getSchemaFor = (dsnpType: DSNPType): Schema => {
  switch(dsnpType) {
    case DSNPType.GraphChange:
      return GraphChangeSchema;
   // ...
  }
}
```

src/core/batch/parquetSchema.ts
```diff
@@ -80,6 +109,39 @@ export const ReactionSchema = {
...
+export const getBloomFilterOptionsFor = (dsnpType: DSNPType): BloomFilterOptions => {
+  const bloomFilterOptions = [
+    GraphChangeBloomFilterOptions,
+    BroadcastBloomFilterOptions,
+    ReplyBloomFilterOptions,
+    ReactionBloomFilterOptions,
+    ProfileBloomFilterOptions,
+  ];
+
+  return bloomFilterOptions[dsnpType - 1];
```
wilwade, line 146: Same here. Using a switch case will give type safety to making sure that every DSNPType has a return value

src/core/store/interface.ts
```diff
@@ -4,6 +4,34 @@ import { NotImplementedError } from "../utilities/errors";
...
+  read(this: unknown, size: number): void;
+  end(chunk: unknown, cb?: () => void): void;
+  end(chunk: unknown, encoding?: string, cb?: () => void): void;
+}
+
+/**
+ * PassThroughCallback: a callback intended for receiving a write stream to create a batch file
+ */
+export interface PassThroughCallback {
+  (stream: ReadWriteStream): void;
```
wilwade, line 32: We await this, so I think it should be a Promise<void>
```suggestion
  (stream: ReadWriteStream): Promise<void>;
```

wilwade (changes requested):
A few notes, but I really like switching to streams. Nice work.

src/core/batch/batch.ts
```diff
@@ -17,17 +20,49 @@ interface BloomFilterData {
...
  *
- * @param events  An array of DSNPEvents to include in the batch file
- * @returns         A batch file object
+ * @param messages An array of DSNPMessage to include in the batch file
+ * @returns         A URL of the storage location
+ * @throws error if messages argument is empty.
  */
-export const createFile = (_events: DSNPType[]): BatchFileObject => {
-  throw NotImplementedError;
+export const createFile = async (targetPath: string, messages: DSNPMessage[]): Promise<URL> => {
```
wilwade, line 37: `targetPath` is missing in the `@param` docs

src/core/batch/batch.ts
```diff
@@ -17,17 +20,49 @@ interface BloomFilterData {
...
+  return putStream(targetPath, async (writeStream: ReadWriteStream) => {
+    await writeBatch(writeStream, schema, messages, { bloomFilters: bloomFilterOptions });
+  });
+};
+
+/**
+ * writeBatch() takes a series of DSNP messages and returns a Batch file
+ * object for publishing.
+ *
+ * @param messages An array of DSNPMessage to include in the batch file
```
wilwade, line 52: Missing some `@param`s here as well. I wonder if we can add a linter for this. I have the same trouble.

src/core/batch/parquetSchema.ts
```diff
@@ -80,6 +109,55 @@ export const ReactionSchema = {
...
+      return BroadcastSchema;
+    case DSNPType.Reply:
+      return ReplySchema;
+    case DSNPType.Reaction:
+      return ReactionSchema;
+    case DSNPType.Profile:
+      return ProfileSchema;
+    default:
+      throw new Error(`Invalid DSNP type: ${dsnpType}`);
+  }
```
wilwade, line 139: By removing the default, TypeScript will throw a build failure for incomplete, but it will not if there is a default. However as I think you were thinking, that doesn't give protection at run time. So I think we just need to move the throw to below the switch:

```suggestion
  }
  throw new Error(`Invalid DSNP type: ${dsnpType}`);
```

src/core/batch/parquetSchema.ts
```diff
@@ -80,6 +109,55 @@ export const ReactionSchema = {
...
+      return BroadcastBloomFilterOptions;
+    case DSNPType.Reply:
+      return ReplyBloomFilterOptions;
+    case DSNPType.Reaction:
+      return ReactionBloomFilterOptions;
+    case DSNPType.Profile:
+      return ProfileBloomFilterOptions;
+    default:
+      throw new Error(`Invalid DSNP type: ${dsnpType}`);
+  }
```
wilwade, line 162:
```suggestion
  }
  throw new Error(`Invalid DSNP type: ${dsnpType}`);
```

src/core/store/interface.ts
```diff
@@ -4,6 +4,34 @@ import { NotImplementedError } from "../utilities/errors";
...
+  write(chunk: unknown, cb?: (error: Error | null | undefined) => void): boolean;
+  end(chunk: unknown, cb?: () => void): void;
+  end(chunk: unknown, encoding?: string, cb?: () => void): void;
+}
+
+/**
+ * ReadStream: a read stream intended reading a batch file
+ */
+export interface ReadStream {
+  read(this: unknown, size: number): void;
```
wilwade, line 23: Is `this` correct? I would have expected `chunk`

src/core/store/interface.ts
```diff
@@ -12,13 +40,27 @@ export type Content = string | Buffer;
 export interface StoreInterface {
   put: (targetPath: string, content: Content) => Promise<URL>;
```
wilwade, line 41: Should we remove `put` and just force usage of the putStream?

wilwade (changes requested):
Last few things. Looking good.

src/core/batch/parquetSchema.ts
wilwade: It is in TypeScript, but in just ol' JS it will not be if someone is using this direct as they can throw any number at it they want to... Although I guess that would just return an undefined. Also, now that I think about it: Is this accessible outside of the sdk?

wilwade (approved):
🚢

PR #64, opened 2021-06-11
get rid of requireGetConfig, use single getters

shannonwells:
Problem
=======
Follow up to #61 , after discussions and looking at the comments
[Finishes #178458945](https://www.pivotaltracker.com/story/show/178458945)

Solution
========
* requireGetConfig is gone.
* Use single getters if you want a guarantee that something is defined or will throw.
* Keeping 'requireGetFoo' and 'getFoo' where applicable as it's an indicator of what is returned; Configs that are not optional fields don't need a 'require' and getters for optional settings if you don't care if they aren't set can be implemented as needed.
* add @returns in typedocs
* tests for all the getters

This eliminates the code smell of using the "as Provider" and friends.
Code is even simpler than before.

Steps to Verify:
----------------
1. Tests pass.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -3,18 +3,21 @@ import {
   getConfig,
   setConfig,
   Config,
-  getQueue,
-  ConfigOpts,
   MissingStore,
+  MissingSigner,
+  MissingProvider,
+  requireGetCurrentFromId,
+  requireGetProvider,
+  requireGetSigner,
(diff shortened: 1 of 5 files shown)


enddynayn (approved):
👍 Nice. I like the simplicity.

aneyzberg (approved):
this looks great!

PR #66, opened 2021-06-15
chore(batch): update reading batch function

enddynayn:
Update #openURL function so that it matches
new interface from ParquetJS API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,12 +109,11 @@
   },
   "dependencies": {
     "@dsnp/contracts": "0.1.0",
-    "@dsnp/parquetjs": "0.9.0",
+    "@dsnp/parquetjs": "1.0.0",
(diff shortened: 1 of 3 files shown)


shannonwells (approved):

package.json
```diff
@@ -113,7 +113,6 @@
     "@ethersproject/abi": "^5.3.0",
     "ethers": "^5.3.0",
     "js-sha3": "^0.8.0",
-    "request": "^2.88.2",
```
wilwade, line 116: Doesn't this require updating the `@dsnp/parquetjs` dependency?

wilwade (changes requested):
One double check

wilwade (approved):
🎉

PR #68, opened 2021-06-15
Naming conventions 178186154

amparks100:
Problem
=======
Variable name usage has not been consistent 
[link to Pivotal Tracker #178186154](https://www.pivotaltracker.com/story/show/178186154)

Solution
========
Updated variable/function names to follow patterns & added linters

with @rosalinekarr


src/core/batch/batch.test.ts
```diff
@@ -1,7 +1,7 @@
-import { batchIncludes } from "./batch";
+import { includes } from "./batch";
```
shannonwells, line 1: was this an eslint recommendation to avoid filename "stuttering"?

src/core/contracts/contract.ts
```diff
@@ -28,7 +29,7 @@ interface ContractResult {
   blockHash: string;
 }
 
-export const DSNPMigrationABI: ethers.utils.ParamType[] = [
+export const DSNP_MIGRATION_ABI: ethers.utils.ParamType[] = [
```
shannonwells, line 32: if we're already changing this file, my preference is to move constant _values_  to the top, between imports and the first exported function.

src/core/contracts/registry.ts
```diff
@@ -33,10 +33,10 @@ export const resolveRegistration = async (handle: Handle, opts?: ConfigOpts): Pr
     const [dsnpUserId, contractAddr] = await contract.resolveRegistration(handle);
     return {
       handle,
-      dsnpUserId: bigNumberToDSNPUserId(dsnpUserId),
+      dsnpUserId: convertBigNumberToDSNPUserId(dsnpUserId),
       contractAddr,
     };
-  } catch (e) {
+  } catch (e: any) {
```
shannonwells, line 39: this should be an Error type

src/core/contracts/registry.ts
```diff
@@ -101,38 +101,38 @@ export const changeHandle = async (
...
+ * @param filter - By dsnpUserId or Contract Address
+ * @param opts - (optional) any config overrides.
  * @returns An array of all the matching events
  */
 export const getDSNPRegistryUpdateEvents = async (
   filter: Partial<Omit<Registration, "handle">>,
   opts?: ConfigOpts
 ): Promise<Registration[]> => {
-  const dsnpUserId = filter.dsnpUserId ? dsnpUserIdToBigNumber(filter.dsnpUserId) : undefined;
+  const dsnpUserId = filter.dsnpUserId ? convertDSNPUserIdToBigNumber(filter.dsnpUserId) : undefined;
```
shannonwells, line 113: I think we can drop 'convert' here; it's common to have simply `thingoneToThingtwo` for conversions.

src/core/utilities/identifiers.ts
```diff
@@ -24,13 +24,13 @@ export const validateDSNPId = (id: string): id is DSNPId => id.match(/dsnp:\/\/[
 export type DSNPUserId = string;
 
 /**
- * bigNumberToDSNPUserId() converts ethers' ridiculous BigNumber implementation
+ * convertBigNumberToDSNPUserId() converts ethers' ridiculous BigNumber implementation
```
shannonwells, line 27: same as above re: `convert`

src/generators/dsnpGenerators.ts
```diff
@@ -70,27 +70,27 @@ export const generateBroadcast = (from?: EthereumAddress): BroadcastMessage => {
...
     contentHash: generateHexString(64),
-    uri: sample(prefabURLs),
+    uri: sample(PREFAB_URLS),
   };
 };
 
 /**
  * generateReaction
- * @param from - a desired fromID (optional)
+ * @param from - - a desired fromID (optional)
```
shannonwells, line 93: two dashes

src/generators/dsnpGenerators.ts
```diff
@@ -70,27 +70,27 @@ export const generateBroadcast = (from?: EthereumAddress): BroadcastMessage => {
...
     contentHash: generateHexString(64),
-    uri: sample(prefabURLs),
+    uri: sample(PREFAB_URLS),
   };
 };
 
 /**
  * generateReply
- * @param from - a desired fromID (optional)
+ * @param from - - a desired fromID (optional)
```
shannonwells, line 79: two dashes

src/search.ts
```diff
@@ -5,29 +5,23 @@ import { NotImplementedError } from "./core/utilities/errors";
 
 type BlockNumber = number;
 
-/**
- * @typedef BaseFilters
```
shannonwells, line 9: typedoc preset doesn't like this?

src/test/hardhatRPC.ts
```diff
@@ -4,7 +4,8 @@ import { ethers } from "ethers";
 
 const RPC_URL = String(process.env.RPC_URL);
 
-const latestSnapshot: string[] = [];
+// eslint-disable-next-line prefer-const
+let latestSnapshot: string[] = [];
```
shannonwells, line 8: I think the const is fine since we're not reassigning to latestSnapshot, we're just messing with its insides?

shannonwells (changes requested):
Just a few nits, and wow this is already a big improvement.

enddynayn:
great improvement! 👍

src/core/batch/batch.test.ts
rosalinekarr: No, we said we wanted to change all methods names to start with a verb. This seemed appropriate since the module was already named "batch." Do you have an alternate suggestion?

src/core/contracts/registry.ts
rosalinekarr: We added this because we wanted all method names to start with a verb. Should we rethink that decision?

src/search.ts
rosalinekarr: yep, interface docs are actually discouraged in general, but if we want to have them, we can use labels on each key.

src/test/hardhatRPC.ts
rosalinekarr: That would mean this is a global const, which would mean it should be named `LATEST_SNAPSHOT` instead. Should we rename it? Or just change it to a const and disable the naming convention instead?

src/core/batch/batch.test.ts
shannonwells: Nah, I was just wondering what the impetus was; I like it, but  the verb thing hadn't occurred to me.

src/core/contracts/registry.ts
shannonwells: yeah, my argument is this might be one exception to make, particularly since it's pretty standard not to have "convert" in front of conversions, especially since it makes this function name really long, OTOH if we started all conversions with "convert" then it makes it really easy to search for conversions, so 🤷‍♀️ .  It's not a blocker, just an idea.

src/test/hardhatRPC.ts
shannonwells: That's an interesting question.  What does "constancy" mean really?  To me the all caps has an implied meaning of not being permuted ever, not even on the inside.  However a `let` implies and allows reassignment entirely.  This is also in test so, maybe a let is fine?

PR #70, opened 2021-06-16
Move S3Node to Examples and Fix up Tests

wilwade:
Problem
=======
We don't want to include S3 as a package

Solution
========
Moved S3Node to an examples directory

with @shannonwells @enddynayn 

Change summary:
---------------
* Moved S3Node to an examples directory
* Updated Store interface to require put and removed get entirely
* Removed store interface redirection and instead used requireGetStore
* Fixed bug with creating parquet filters
* Update Test Store to do a better job testing and catching parquetjs bugs


src/content.ts
```diff
@@ -42,7 +43,10 @@ export const broadcast = async (
   const currentFromId = config.requireGetCurrentFromId(opts);
 
   const contentHash = keccak256(content);
-  const uri = await store.put(contentHash, content, opts);
+  const store = requireGetStore(opts);
+  const uri = await store.putStream(contentHash, async (stream: WriteStream) => {
```
enddynayn, line 47:
```
stream.write(Buffer.from(content));
stream.end();
```

src/content.ts
```diff
@@ -74,7 +78,10 @@ export const reply = async (
...
+    stream.end(content);
```
enddynayn, line 83: stream.write(Buffer.from(content));
stream.end();

src/content.ts
```diff
@@ -74,7 +78,10 @@ export const reply = async (
   const currentFromId = config.requireGetCurrentFromId(opts);
 
   const contentHash = keccak256(content);
-  const uri = await store.put(contentHash, content, opts);
+  const store = requireGetStore(opts);
+  const uri = await store.putStream(contentHash, async (stream: WriteStream) => {
```
enddynayn, line 82: randomFilename
uuid?
`uuid-timestamp-contentHash`

src/content.ts
```diff
@@ -126,7 +133,10 @@ export const profile = async (
...
+    stream.end(content);
```
enddynayn, line 138:
```
stream.write(Buffer.from(content));
stream.end();
```

src/core/batch/batch.ts
```diff
@@ -41,7 +42,8 @@ export const createFile = async (targetPath: string, messages: DSNPMessage[]): P
   const schema = new ParquetSchema(getSchemaFor(messages[0].dsnpType));
   const bloomFilterOptions = getBloomFilterOptionsFor(messages[0].dsnpType);
 
-  return putStream(targetPath, async (writeStream: ReadWriteStream) => {
+  const store = requireGetStore();
```
enddynayn, line 45: update test for this

src/core/store/interface.ts
```diff
@@ -29,7 +26,7 @@ export interface ReadStream {
  * PassThroughCallback: a callback intended for receiving a write stream to create a batch file
  */
 export interface PassThroughCallback {
-  (stream: ReadWriteStream): Promise<void>;
+  (stream: WriteStream): Promise<void>;
```
enddynayn, line 29: undo to `(stream: ReadWriteStream): Promise<void>;`

src/core/store/interface.ts
```diff
@@ -38,50 +35,31 @@ export interface PassThroughCallback {
...
+   */
   put?: (targetPath: string, content: Content) => Promise<URL>;
+
+  /**
+   * get() fetches a batch file from the given target path.
+   *
+   * @param   targetPath  The path to and name of file.
+   * @returns   The batch file fetched from the given URI
+   */
   get?: (targetPath: string) => Promise<File>;
```
enddynayn, line 54: discuss removing. Discuss having different interface which contains union of other interfaces.

src/content.ts
```diff
@@ -42,7 +43,10 @@ export const broadcast = async (
   const currentFromId = config.requireGetCurrentFromId(opts);
 
   const contentHash = keccak256(content);
-  const uri = await store.put(contentHash, content, opts);
```
enddynayn, line 45: is contentHash a good enough file name? Could we run into problems when sending the same note. At scale, even a timestamp might not be enough to differentiate it. Perhaps, UUID might be a good idea.

wilwade: I think content hash is. If we send the same note twice with the same contents, I think it is ok if they use the same file.

src/core/store/interface.ts
wilwade: I'm up for removing it

src/content.ts
wilwade: Ended up just keeping put interface

src/content.ts
wilwade: Let's discuss, but not change in this PR.

src/core/store/interface.ts
wilwade: @enddynayn We currently only use the WriteStream part of this. Is this needed for the Read part? (I'll go check #66)

src/core/store/interface.ts
wilwade: Removed.

src/core/batch/batch.test.ts
```diff
@@ -1,100 +1,101 @@
...
-  const messages = [generateBroadcast()];
+    it("calls ParquetWriter#appendRow to add a row to parquet stream", async () => {
+      await writeBatch(writeStream, BroadcastSchema, messages);
+      expect(parquetWriterInstance.appendRow).toHaveBeenCalledTimes(1);
+    });
 
-  beforeEach(() => {
-    jest.spyOn(batch, "writeBatch");
-    (putStream as jest.Mock).mockImplementation((path, callback) => callback(path, callback));
+    it("calls ParquetWriter#closes to end the stream", async () => {
```
shannonwells, line 64: minor typo, I say only fix if you're going to make another change in this file.

shannonwells (approved):
This looks ready to go. Nothing blocking.

examples/s3Node.ts
```diff
@@ -1,23 +1,8 @@
-import { S3Client, S3ClientConfig, GetObjectCommand, PutObjectCommand } from "@aws-sdk/client-s3";
-import { StoreInterface } from "./interface";
+// import { PassThroughCallback, StoreInterface } from "@dsnp/sdk/core/store";
```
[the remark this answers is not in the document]
wilwade: This is actually an example comment. It is showing how to import the StoreInterface from the *real* sdk. Not sure if it is needed however.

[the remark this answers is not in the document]
shannonwells: If that's so then I think it needs some explanation, since it just looks like you accidentally left it in.  It's better to put it in maybe a README in this directory though, along with some other info maybe?

enddynayn (approved):

PR #71, opened 2021-06-16
Changed setConfig to preserve existing keys in the config when called

rosalinekarr:
Problem
=======
We would prefer if the `setConfig()` method merged new config options with existing ones to avoid losing settings when called.
[#178559063](https://www.pivotaltracker.com/story/show/178559063)

Solution
========
Update it to do that

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -72,7 +72,10 @@ export const getConfig = (overrides?: ConfigOpts): Config => {
 };
 
 /**
- * setConfig() sets the current configuration with the given object.
+ * setConfig() sets the current configuration with the given object. Any keys
+ * previously set on the config object will not be removed. To remove a config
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -81,9 +81,7 @@ export const setConfig = (newConfig: ConfigOpts): Config => {
   const { signer, provider } = newConfig;
```
wilwade, line 81: @rosalinekarr could you add some clear text in the documentation for this method to note that it does this and how to remove options?

wilwade (changes requested):
Documentation request

src/config.ts
rosalinekarr: fixed

wilwade (approved):
setConfig({approve: true});

amparks100 (approved):

PR #72, opened 2021-06-16
Removed autheticateHandle method and updated follow and unfollow methods

rosalinekarr:
Problem
=======
We don't want to use handles in any methods exposed to the user except `resolveHandle()`.
[#178559050](https://www.pivotaltracker.com/story/show/178559050)

Solution
========
Replace handles with DSNPUserIds in all methods using them and delete `authenticateHandle()`

### DIFF
--- a/src/handles.test.ts
+++ b/src/handles.test.ts
@@ -1,14 +1,6 @@
 import { Signer } from "ethers";
 import * as config from "./config";
-import {
-  authenticateHandle,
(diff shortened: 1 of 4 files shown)


wilwade (approved):

enddynayn (approved):

enddynayn:
This a breaking change. 1.0.0?

PR #73, opened 2021-06-18
Removed S3 from the SDK

rosalinekarr:
Problem
=======
We want the storage system used for the SDK to be modular so implementers may choose to any hosting service they prefer to store their content and batch files, but currently, S3 is a hard dependency in the SDK.

Solution
========
Remove S3 from the SDK and update the README to point at some newly created adapter repos.


rosalinekarr:
See also:
https://github.com/rosalinekarr/dsnp-google-store
https://github.com/rosalinekarr/dsnp-s3-store

wilwade (changes requested):
We'll need to reconcile this with #70

rosalinekarr:
Ah, oops. I didn't realize there was already a PR for this. Yeah, we can reconcile them on Monday.

PR #75, opened 2021-06-21
Features/sdk announce stream 178308136

aneyzberg:
Problem
=======
We need a way to be able to consistently listen for and retrieve batchAnnounce events that happen on chain
[link to Pivotal Tracker #178308136](https://www.pivotaltracker.com/story/show/178308136)

Solution
========
We added a function that sets up a listener to listen for and returns batchAnnounce events from the chain. 
This function takes a callback and a filter - and will filter through events and then invoke the callback for each filtered event.

with @wilwade 

Change summary:
---------------
* Add a new `subscription.ts` module
* Add functionality to set up listener in function `batchAnnounceEvents`
* Add functionality to filter events in `batchAnnounceEvents`
* have `batchAnnounceEvents deal with retrieving both bast and current events`
* Add spec
* Add `utilities.ts` file in test with a `checkNumberOfFunctionCalls` function that waits for our callback to get called
* use new ganache container in circle so we no longer need to look for the contracts and deploy them to a container.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,12 +9,9 @@ jobs:
     runs-on: ubuntu-20.04
     services:
       ganache:
-        image: unfinishedlabs/ganache-truffle-node:latest
+        image: dsnp/ganache:v0.1.0
         ports:
           - 8545:8545
-        options: >-
-          -e MNEMONIC="test test test test test test test test test test test junk"
-          -e CHAINID=31337
     steps:
       - name: Checkout SDK
         uses: actions/checkout@v2
@@ -39,41 +36,6 @@ jobs:
       - name: Install
         run: npm ci
         working-directory: ./main
-
-      - name: Find contracts package in SDK
-        id: contracts_version
-        run: |
-          npm list @dsnp/contracts
-          export version=`npm list @dsnp/contracts --silent | sed -En 's/.*contracts@0.0.0-(.{6}).*$/\1/p'`
-          export version=`[ "$version" ] || npm list @dsnp/contracts --silent | sed -En 's/.*contracts@(.*)$/v\1/p'`
-          echo "Found version ${version}"
-          git clone --bare https://github.com/LibertyDSNP/contracts.git contracts-temp
-          cd contracts-temp
-          export result=`git rev-parse $version`
-          [ "$result" ] || (echo "Unable to find the right git SHA for ${version}" && exit 1)
-          echo "Using SHA ${result}"
-          cd ..
-          rm -Rf ./contracts-temp
-          echo "contract_version=$result" >> $GITHUB_ENV
-        shell: bash
-        working-directory: ./main
-
-      - name: Checkout Contracts
-        uses: actions/checkout@v2
-        with:
-          repository: LibertyDSNP/contracts
-          path: contracts
-          ref: ${{ env.contract_version }}
-
-      - name: Install
-        run: npm ci
-        working-directory: ./contracts
-
-      - name: deploy contracts
-        run:  npm install hardhat && npm run build && npm run deploy:localhost
-        working-directory: ./contracts
-        env:
-          LOCAL_NETWORK_ACCOUNT_PRIVATE_KEY: ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 
       - name: Lint
         run: npm run lint
--- a/src/core/contracts/announcement.ts
+++ b/src/core/contracts/announcement.ts
@@ -3,7 +3,6 @@ import { ConfigOpts, requireGetProvider, MissingContract, getContracts, requireG
 import { HexString } from "../../types/Strings";
 import { Announcer, Announcer__factory } from "../../types/typechain";
 import { getContractAddress } from "./contract";
-// import { LogDescription } from "@ethersproject/abi/src.ts/interface";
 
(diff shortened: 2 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -9,12 +9,9 @@ jobs:
     runs-on: ubuntu-20.04
     services:
       ganache:
-        image: unfinishedlabs/ganache-truffle-node:latest
+        image: dsnp/ganache:v0.1.0
```
wilwade, line 12: 🎉

src/core/contracts/announcement.ts
```diff
@@ -1,9 +1,9 @@
-import { ContractTransaction, ethers, EventFilter } from "ethers";
+import { ContractTransaction, EventFilter } from "ethers";
 import { ConfigOpts, requireGetProvider, MissingContract, getContracts, requireGetSigner } from "../../config";
 import { HexString } from "../../types/Strings";
-import { abi as announcerABI } from "@dsnp/contracts/abi/Announcer.json";
 import { Announcer, Announcer__factory } from "../../types/typechain";
 import { getContractAddress } from "./contract";
+// import { LogDescription } from "@ethersproject/abi/src.ts/interface";
```
wilwade, line 6: comment

src/core/contracts/subscription.test.ts
```diff
@@ -0,0 +1,174 @@
...
+import { requireGetProvider } from "../../config";
+import { checkNumberOfFunctionCalls } from "../../test/utilities";
+import { BatchAnnounceCallbackArgs } from "./subscription";
+
+describe("subscription", () => {
+  setupSnapshot();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.resetAllMocks();
```
wilwade, line 16: reset does everything clear does (+ extra)
```suggestion
    jest.resetAllMocks();
```

Actually, you don't reuse any mocks, so:

```suggestion
```

src/core/contracts/subscription.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    setupConfig();
+  });
+
+  afterEach(async () => {
+    jest.clearAllMocks();
+    jest.resetAllMocks();
```
wilwade, line 22:
```suggestion
```

src/core/utilities/index.ts
```diff
@@ -4,17 +4,17 @@
  * export * from "./foo"
  * ```
  * Leads to use like:
- *   - `import { functionInFoo } from "@dsnp/sdk/utilities";`
- *   - `import util from "@dsnp/sdk/utilities"; util.functionInFoo();`
+ *   - `import { functionInFoo } from "@dsnp/sdk/utilities.ts";`
+ *   - `import util from "@dsnp/sdk/utilities.ts"; util.functionInFoo();`
```
wilwade, line 8: Not sure why these changed.

src/core/utilities/index.ts
```diff
@@ -4,17 +4,17 @@
...
  * To export at a nested level:
  * ```
  * import * as fooImport from "./foo";
  * export const foo = fooImport;
  * ```
  * Leads to use like:
- *   - `import util from "@dsnp/sdk/utilities"; util.foo.functionInFoo();`
- *   - `import { foo } from "@dsnp/sdk/utilities"; foo.functionInFoo();`
+ *   - `import util from "@dsnp/sdk/utilities.ts"; util.foo.functionInFoo();`
+ *   - `import { foo } from "@dsnp/sdk/utilities.ts"; foo.functionInFoo();`
```
wilwade, line 17: Same

wilwade (dismissed):
Few small things

src/core/contracts/subscription.test.ts
aneyzberg: oh yes. This I meant to clean up a while ago.

src/core/utilities/index.ts
aneyzberg: I am not sure either.

wilwade (commented):
lgtm. We paired some on it, so be nice to get another set of eyes on it before merging.

src/core/contracts/subscription.ts
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * batchAnnounceEvents() sets up a listener to listen to retrieve Batch Announce events from the chain.
+ * It takes a callback and a filter. The filter is used to filter events that come through.
+ * The callback is invoked for each correctly filtered event.
+ *
+ * @param callback - The callback function to be called when an event is received
+ * @param filters -  Any filter options for including or excluding certain events
+ * @returns        A function that can be called to remove listener for this type of event
+ */
+export const batchAnnounceEvents = async (
```
rosalinekarr, line 40: Can we rename this to something like `subscribeToBatchAnnounceEvents` or just `subscribe`? This naming makes it seem like this method is for announcing a batch.

rosalinekarr (approved):
LGTM, only non-blocking feedback

src/core/contracts/subscription.ts
rosalinekarr: Also, depending on whether "batch" is a verb or a noun here, this may be breaking our function naming convention.

aneyzberg: It is in the subscription module. But I do see your point. I can do that.

PR #81, opened 2021-06-30
Features/commit 178625369

rosalinekarr:
Problem
=======
Users need a way to take all the DSNP messages they've queued up for publishing and post them on chain
[#178625369](https://www.pivotaltracker.com/story/show/178625369)

Solution
========
Add a porcelain function that takes an iterable of DSNP messages, stores them and returns an array of Announcements


src/core/batch/batch.ts
```diff
@@ -20,47 +23,73 @@ interface BloomFilterData {
   RowGroupIndex: number;
 }
 
+interface BatchFileData {
+  url: URL;
+  hash: HexString;
+}
+
+type BatchMessage<T extends DSNPType> = AsyncOrSyncIterable<DSNPMessageSigned<DSNPTypedMessage<T>>>;
```
wilwade, line 31: We might need to pluralize this at least. It reads oddly down below when we replaced `DSNPMessage[]` with `BatchMessage<T>`

Perhaps `BatchMessagesIterable` event?

src/core/utilities/index.ts
```diff
@@ -19,3 +19,5 @@
 
 export * from "./errors";
 export * from "./identifiers";
+export * from "./iterator";
+export * from "./random";
```
wilwade, line 23: Random doesn't feel like something we should be exporting outside of the sdk. Not sure how to handle that however. It quickly gets confusing of some things in utilities are and some are not.

wilwade (commented):
Two small notes for the draft unrelated to the work on iterables.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "esnext",
```
wilwade, line 3: Do we still need this change?

rosalinekarr: yep, we fail to compile without it

rosalinekarr: Changed `tsconfig.json` target back to `es6` and updated `tsconfig.eslint.json` to target `esnext` so it doesn't complain about `AsyncIterable` types.

src/core/batch/batch.ts
```diff
@@ -20,47 +23,115 @@ interface BloomFilterData {
...
+ * @returns A BatchFileData object including a URL and keccak hash of the file
  */
-export const createFile = async (targetPath: string, messages: DSNPMessage[], opts?: ConfigOpts): Promise<URL> => {
-  if (messages.length === 0) throw EmptyArrayError;
-
-  const schema = new ParquetSchema(getSchemaFor(messages[0].dsnpType));
-  const bloomFilterOptions = getBloomFilterOptionsFor(messages[0].dsnpType);
+export const createFile = async <T extends DSNPType>(
+  targetPath: string,
+  dsnpType: T,
```
enddynayn, line 82: We can resolve the type from the BatchIterable why do we also pass it in?

src/core/batch/batch.ts
```diff
@@ -20,47 +23,115 @@ interface BloomFilterData {
...
+ * stream and returns a void promise which resolves when done.
  *
  * @param writeStream - A writable stream
  * @param schema - The ParquetJS schema for the messages DSNP type
  * @param messages - An array of DSNPMessage to include in the batch file
  * @param opts - Options for creating a Parquet file
  * @returns A void promise which will either resolve or reject
  */
-export const writeBatch = async (
+export const writeBatch = async <T extends DSNPType>(
```
enddynayn, line 119: DSNPType is an enum can you please explain what this does? I am still a newbie at Typescript.
Why not just use DSNPType?

src/createAnnouncements.ts
```diff
@@ -0,0 +1,61 @@
...
+ * createAnnouncement takes a DSNP type and an array of messages of the given
+ * type, generates a batch file from the messages, stores them and returns an
+ * annoucement linking to the file.
+ *
+ * @param dsnpType - The DSNPType of the messages passed in
+ * @param messages - The DSNPBatchMessages to publish
+ * @param opts - Optional. Configuration overrides, such as from address, if any
+ * @returns A promise of the generated annoucement
+ */
+export const createAnnoucement = async <T extends DSNPType>(
```
enddynayn, line 19: such an Enddy move with misspelling createAnnoucement

src/createAnnouncements.ts
```diff
@@ -0,0 +1,61 @@
...
+  messages: AsyncOrSyncIterable<DSNPBatchMessage>,
+  opts?: ConfigOpts
+): Promise<Announcement[]> => {
+  const announcements: Record<string, Promise<Announcement>> = {};
+
+  for await (const message of messages) {
+    const dsnpType = message.dsnpType;
+
+    if (!announcements[dsnpType]) {
+      const filteredMessages = filterIterable<DSNPBatchMessage>(messages, (message) => message.dsnpType == dsnpType);
```
enddynayn, line 54: what about a groupByIterable instead?

enddynayn (approved):

src/core/batch/batch.ts
wilwade: This generic allows us to get specific about which type of BatchIterable we have. Although in this case we don't need it because only part of each message we use is `msg.dsnpType`. So we could do this instead I think:

```suggestion
export const writeBatch = async (
  writeStream: WriteStream,
  schema: Schema,
  messages: BatchIterable<DSNPType>,
```

src/core/batch/batch.ts
```diff
@@ -20,47 +23,115 @@ interface BloomFilterData {
...
  */
-export const createFile = async (targetPath: string, messages: DSNPMessage[], opts?: ConfigOpts): Promise<URL> => {
-  if (messages.length === 0) throw EmptyArrayError;
-
-  const schema = new ParquetSchema(getSchemaFor(messages[0].dsnpType));
-  const bloomFilterOptions = getBloomFilterOptionsFor(messages[0].dsnpType);
+export const createFile = async <T extends DSNPType>(
+  targetPath: string,
+  dsnpType: T,
+  messages: BatchIterable<T>,
```
wilwade, line 83: Due to Enddy's comment below, I think we can also remove this generic and just do:
```suggestion
  messages: BatchIterable<DSNPType>,
```

Why? Because we don't need it to type the return more specifically and the only field we use is `dsnpType`.

src/core/batch/batch.ts
rosalinekarr: My intention with this generic was to add a type check that messages iterable passed to this function only ever has one DSNPType, i.e. that it's something like `[Broadcast, Broadcast, Broadcast]` not `[Broadcast, Reply, Broadcast]`.

I know this check only applies if the user is also using TypeScript, so I also added a logical check in the body of the function below which throws `MixedTypeBatchError` if multiple are passed, but my thinking was that *if* the SDK user is also using TypeScript, they can avoid that error entirely with some extra type safety.

Do we want to drop this generic and lose that type safety?

src/createAnnouncements.ts
rosalinekarr: A groupBy function would be fairly complicated to implement in an asynchronous way. If you can think of a way to implement one, I'm happy to change it.

src/core/batch/batch.ts
rosalinekarr: See earlier reply: https://github.com/LibertyDSNP/sdk-ts/pull/81#discussion_r666197951

src/core/batch/batch.ts
wilwade: That is a really good point. I had forgotten about that type check.

Keep the safety!

src/core/batch/batch.ts
rosalinekarr: fixed

src/createAnnouncements.ts
rosalinekarr: fixed

src/createAnnouncements.ts
enddynayn: why does the iterator need to be asynchronous?

rosalinekarr: So we can write multiple batch files at the same time. I believe it was @wilwade's intention that the entire batch writing process be streaming.

wilwade (approved):

PR #83, opened 2021-07-01
feat(identity): add non-EIP712 delegate method

enddynayn:
Add ability to add or update a delegate.

### DIFF
--- a/src/core/contracts/identity.test.ts
+++ b/src/core/contracts/identity.test.ts
@@ -9,20 +9,25 @@ import {
   createAndRegisterBeaconProxy,
   isAuthorizedTo,
   Permission,
+  upsertDelegate,
 } from "./identity";
 import { EthAddressRegex } from "../../test/matchers";
 import { setupConfig } from "../../test/sdkTestConfig";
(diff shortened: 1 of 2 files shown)


src/core/contracts/identity.test.ts
```diff
@@ -137,6 +142,24 @@ describe("identity", () => {
...
+      contractOwner = await signer.getAddress();
+      const identityContract = await new Identity__factory(signer).deploy(contractOwner);
+      await identityContract.deployed();
+      contractAddress = identityContract.address;
+    });
+
+    it("upserts delegate", async () => {
+      await upsertDelegate(contractAddress, NON_OWNER, 0x1);
+
+      expect(await isAuthorizedTo(NON_OWNER, contractAddress, 2, 60)).toBeTruthy();
```
acruikshank, line 160: This should really have an update as well as an insert test. It would be ok to add a second call to this test with a second expectation.

acruikshank (approved):
This looks good. A little more testing would be nice, but nothing blocking.

shannonwells (approved):
LGTM other than @acruikshank 's comment about the testing.

PR #84, opened 2021-07-02
chore(CI): swap Ganache for Harhat

enddynayn:
Swap Ganache for Hardhat. Version 6 of
Ganache has the chainid opcode hardcoded to 1. 
As a result, it prevents EIP-712 signatures from being valid.
The EIP-712 domain separator includes a chain-id that
will cause a discrepancy when signing and verifying
on-chain.


wilwade (approved):
🎩

PR #86, opened 2021-07-10
Chores/better errors 178488570

rosalinekarr:
The goal of this chore was actually to put together a list of potential improvements for discussion and story writing, but in the process of looking at how we currently handle errors in the SDK, I found a bunch of little, obvious improvements that we could go ahead and make, so I started a branch, made fixes as I saw them and everything sort of snowballed from there. Sorry for the giant diff ☹️

I'll have a list of further improvements we can make for the next Spec/SDK meeting.
[#178488570](https://www.pivotaltracker.com/story/show/178488570)

Change summary:
---------------
* Replaced all our text errors with proper error classes.
* Fixed a couple places where we were actually returning the wrong error for certain failures.
* Cleaned up the way we handle missing config errors.
* Changed all our error tests to be consistent.
* Updated all the method docs to include errors thrown.


src/config.ts
```diff
@@ -129,16 +114,17 @@ export const requireGetStore = (opts?: ConfigOpts): StoreInterface => {
...
 export const requireGetCurrentFromId = (opts?: ConfigOpts): string => {
   const c = getConfig(opts);
-  if (!c.currentFromId) throw MissingUser;
+  if (!c.currentFromId) throw new MissingFromIdConfigError();
   return c.currentFromId;
 };
 
 /**
- * Get the contracts.  Since this is a required field, this is a plain getter.
+ * Get the contracts
```
rosalinekarr, line 122: I think there was some confusion about this config option vs the `getContractAddress` method in the contracts module. The Contract config option is actually never required and there are no defaults. It's only intended to be an override for user who would rather use their own solidity contracts instead of ours. Otherwise, we fetch our current contract ABI from the chain, not the config.

At least, that's my understanding. Please, let me know if I have it wrong.

src/core/config/configErrors.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * MissingFromIdConfigError indicates that the current from id is missing.
+ */
+export class MissingFromIdConfigError extends ConfigError {
+  constructor() {
+    super("No from id found. Please authenticate a handle.");
+    this.name = "MissingUFromIdConfigError";
+  }
+}
```
rosalinekarr, line 52: We can discuss it more in the next SDK meeting, but I was thinking, since `setConfig` is the only config module method we actually use at the porcelain level, we might want to move the rest of the config back into a plumbing module here with these errors, and expose a separate config porcelain with the one method we actually expect the users to call.

src/core/contracts/identity.ts
```diff
@@ -128,40 +141,34 @@ export const createAndRegisterBeaconProxy = async (
...
   return IdentityCloneFactory__factory.connect(address, signer);
 };
 
 const getBeaconFactoryContract = async (opts?: ConfigOpts): Promise<BeaconFactory> => {
-  const { beaconFactory } = getContracts(opts);
   const signer = requireGetSigner(opts);
   const provider = requireGetProvider(opts);
-
-  const address = beaconFactory || (await getContractAddress(provider, BEACON_FACTORY_CONTRACT));
-  if (!address) throw MissingContract;
```
rosalinekarr, line 157: So this pattern turns out to be really confusing for the end user. The `MissingContract` error message tells the user that they need to set their contract addresses in the config, but it's thrown here when we can't find our own contracts on the chain. We were doing this all over the contracts module.

I think we should make a point to be more clear about error naming in the future, so we don't wind up throwing the wrong error message for a problem again.

src/config.ts
```diff
@@ -1,29 +1,14 @@
...
 import { StoreInterface } from "./core/store";
 import { HexString } from "./types/Strings";
 
-export const MissingContract = new Error("Contract was not found");
-export const MissingSigner = new Error("Signer is not set.");
-export const MissingProvider = new Error("Blockchain provider is not set.");
-export const MissingStore = new Error("Store adapter was not found");
-export const MissingUser = new Error("No user id found. Please authenticate a handle.");
-
-export interface Contracts {
```
wilwade, line 12: I don't understand the reason for making this generic. It loses specificity and documentation right?

src/content.ts
```diff
@@ -55,21 +57,31 @@ export const broadcast = async (
...
+ * @throws {@link MissingSignerConfigError}
+ * Thrown if the signer is not configured.
+ * @throws {@link MissingStoreConfigError}
+ * Thrown if the store is not configured.
+ * @throws {@link MissingFromIdConfigError}
+ * Thrown if the from id is not configured.
+ * @throws {@link InvalidActivityPubError}
+ * Thrown if the provided activity pub object is not valid.
+ * @throws {@link InvalidMessageIdentifierError}
+ * Thrown if the provided inReplyTo DSNP Message Id is invalid.
```
wilwade, line 69: 🙇

src/core/config/configErrors.ts
wilwade: I'm not sure about where inside, but I like the idea of only externally exposing `getConfig` and `setConfig` (aka never exposing the other *Config methods outside of the sdk)

src/core/config/configErrors.ts
```diff
@@ -0,0 +1,52 @@
+import { DSNPError } from "../errors";
+
+/**
+ * ConfigError indicates that some piece of configuration is not set and that
+ * the current action cannot proceed without it.
+ */
+export class ConfigError extends DSNPError {
```
wilwade, line 7: I don't see `core/config/configErrors` getting exported out of the sdk, but I think we would want them to be?

src/core/contracts/contractErrors.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * NoLogsFoundContractError indicates that a log event could not be found.
+ */
+export class NoLogsFoundContractError extends ContractError {
+  eventName: string;
+
+  constructor(eventName: string) {
+    super(`Could not find log event: ${eventName}.`);
+    this.name = "NoLogsFoundContractError";
```
wilwade, line 50: Could we introspect this at a lower level?

wilwade (changes requested):
Looks nice. Questions more than changes.

src/config.ts
rosalinekarr: The problem with the interface is that TypeScript won't let us index the object with `[]`, since the string may or may not be a valid key on the object at runtime. I did some investigation, and [what I found](https://stackoverflow.com/questions/57350092/string-cant-be-used-to-index-type) indicated that it's actually bad practice to define interfaces which will be accessed dynamically since the interface can't provide any real type safety, just the illusion of it.

An alternative we could try if we really want to preserve the typing here would be to replace the `getContracts()` method with a series of methods for each specific contract, i.e. `getAnnouncerContract()`, `getRegistryContract()`, etc, but I figured that was too big of a change for this PR.

src/core/config/configErrors.ts
rosalinekarr: Fixed

src/core/contracts/contractErrors.ts
rosalinekarr: Fixed

src/config.ts
rosalinekarr: **Notes from async discussion:**

- We decided to replace this interface with a series of string types for the various contract names and update the contracts type to use a union of the types.
- Fix has already been pushed.

src/config.ts
```diff
@@ -35,7 +41,7 @@ export interface Config {
   /** The Storage handles storing batch, content, and other media files at a publicly accessible location */
   store?: StoreInterface;
   /** Contracts are different addresses for specific contracts or for running custom tests */
-  contracts: Contracts;
+  contracts: { [key in ContractName]?: HexString };
```
wilwade, line 44: We should likely alias this type so we can use it in `getContracts` below.

src/config.ts
```diff
@@ -129,16 +135,17 @@ export const requireGetStore = (opts?: ConfigOpts): StoreInterface => {
...
 
 /**
- * Get the contracts.  Since this is a required field, this is a plain getter.
+ * Get the contracts
  *
  * @param opts - overrides for the current configuration.
+ * @returns potentially undefined contract addresses
  */
-export const getContracts = (opts?: ConfigOpts): Contracts => {
+export const getContracts = (opts?: ConfigOpts): Record<string, HexString> => {
```
wilwade, line 148: Should use the new type.

src/core/contracts/contract.ts
```diff
@@ -73,16 +76,30 @@ const filterValues = (values: ContractResult[], contractName: string): ContractR
...
  * @param contractName - Name of contract to find address for
- * @returns HexString A hexadecimal string representing the contract address
+ * @param opts - Optional. Configuration overrides, such as from address, if any
+ * @returns HexString A hexidecimal string representing the contract address
  */
 export const getContractAddress = async (
   provider: ethers.providers.Provider,
-  contractName: string
-): Promise<HexString | null> => {
+  contractName: string,
```
wilwade, line 97: This should also use the contract names union type.

src/core/contracts/contractErrors.ts
```diff
@@ -0,0 +1,49 @@
...
+    super(`Could not find address for ${contractName} contract.`);
+    this.contractName = contractName;
+  }
+}
+
+/**
+ * MissingRegistrationContractError indicates that the registration for a given
+ * DSNP User Id could not be found.
+ */
+export class MissingRegistrationContractError extends ContractError {
```
wilwade, line 30: I was confused by this name. Perhaps drop `contract`?
```suggestion
export class MissingRegistrationError extends ContractError {
```

src/core/contracts/registry.ts
```diff
@@ -145,8 +172,14 @@ export const isMessageSignatureAuthorizedTo = async (
...
 ): Promise<boolean> => {
-  const reg = await resolveId(dsnpUserId);
-  if (!reg) throw MissingContract;
+  const registrations = await getDSNPRegistryUpdateEvents(
+    {
+      dsnpUserId,
+    },
+    opts
+  );
+  if (registrations.length === 0) throw new MissingRegistrationContractError(dsnpUserId);
```
wilwade, line 181: Do we want to throw in this case? I think just returning false would suffice.

src/core/contracts/subscription.ts
```diff
@@ -29,10 +29,12 @@ export interface BatchFilterOptions {
 type BatchPublicationCallback = (doReceivePublication: BatchPublicationCallbackArgs) => void;
 
 /**
- * subscribeToBatchPublications() sets up a listener to listen to retrieve Batch Announce events from the chain.
+ * subscribeToBatchPublications() sets up a listener to listen to retrieve Batch Publish events from the chain.
```
wilwade, line 32:
```suggestion
 * subscribeToBatchPublications() sets up a listener to listen to retrieve Batch Publication events from the chain.
```

wilwade (changes requested):
A few final small things.

src/core/contracts/registry.ts
rosalinekarr: I'll change it in my current task: https://www.pivotaltracker.com/story/show/177926772

wilwade (approved):
🚂

PR #92, opened 2021-07-14
Features/validate 177926772

rosalinekarr:
Problem
=======
We don't have a complete way to validate announcements anywhere in the SDK yet, only individual pieces of validation like activity content validation and message signatures, and several of those validators are out of date with the spec.
[#177926772](https://www.pivotaltracker.com/story/show/177926772)

Solution
========
Update all our existing validators to comply with the spec and add validation methods to the ~messages~ announcement module which calls them all in one place.


src/content.ts
```diff
@@ -68,35 +68,34 @@ export const broadcast = async (
...
- * @throws {@link InvalidActivityPubError}
- * Thrown if the provided activity pub object is not valid.
+ * @throws {@link InvalidActivityContentError}
+ * Thrown if the provided activity content object is not valid.
  * @throws {@link InvalidAnnouncementIdentifierError}
  * Thrown if the provided inReplyTo Announcement Id is invalid.
- * @param contentOptions - Options for the activity pub content to generate
- * @param inReplyTo - The DSNP Announcement Id of the announcement that this announcement is in reply to
+ * @param contentObject - The activity content object with which to reply
+ * @param inReplyTo - The DSNP Message Id of the message that this message is in reply to
```
wilwade, line 76:
```suggestion
 * @param inReplyTo - The DSNP Announcement Id that this announcement is in reply to
```

src/core/announcements/validation.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  let fileUrl: string | null = null;
+  if (typeof activityContent["url"] === "string") {
+    fileUrl = activityContent["url"];
+  } else {
+    fileUrl = activityContent["url"]["href"];
+  }
+
+  const fileContents = await fetch(fileUrl).then((res) => res.text());
+  return keccak256(fileContents) == keccakHash;
```
wilwade, line 212: I don't think we should be fetching these files. They could be very large, and if they are desired, they should be fetched and cached.

Instead we should expose the hash function (instead of requiring people import js-sha3) (We have a PT story for that: https://www.pivotaltracker.com/story/show/178904143)

src/core/identifiers/identifiers.ts
```diff
@@ -24,6 +27,17 @@ export const validateDSNPAnnouncementId = (id: string): id is DSNPAnnouncementId
...
 
+/**
+ * isDSNPUserId validates a given object as a DSNPUserId.
+ *
+ * @param id - The object to validate
+ * @returns True of false depending on whether the string is a valid DSNPUserId
+ */
+export const isDSNPUserId = (id: unknown): id is DSNPUserId => {
+  if (!isString(id)) return false;
+  return id.match(/^dsnp:\/\/[0-9A-F]{16}$/i) !== null;
```
wilwade, line 38: I'm working on this part of the spec, and a DSNP User Id right now does not have the `dsnp://` prefix

This is so that `fromId` and `objectId` don't need to have that prefix, but can still be referred and validated as DSNP User Ids.

src/core/utilities/random.ts
```diff
@@ -19,3 +22,23 @@ export const getRandomString = (length?: number): string => {
...
+ */
+export const getRandomHex = (length?: number): HexString => {
+  let string = "";
+
+  for (let i = 0; i < (length || 32); i++) {
+    const index = Math.floor(Math.random() * HEX.length);
+    string += HEX[index];
+  }
+
+  return `0x${string}`;
```
wilwade, line 43: Side note: We need to be better about what requires and what doesn't require a `0x` prefix....

src/core/identifiers/identifiers.ts
rosalinekarr: I thought the plan was to use binary in the parquet files and contracts and to immediately translate it to `dsnp://` hex strings everywhere in the SDK for user friendliness. Did that plan change?

src/core/announcements/validation.ts
```diff
@@ -0,0 +1,206 @@
...
+ * linked activity content. It **does not** validate any audio, image or video
+ * files linked in the activity content object, so these files should be
+ * validated separately.
+ *
+ * @param obj - A signed announcement to validate
+ * @param opts - Optional. Configuration overrides, such as from address, if any
+ * @returns True if the announcement is valid, otherwise false
+ */
+export const isValidAnnouncement = async (obj: unknown, opts?: ConfigOpts): Promise<boolean> => {
+  const fetch = requireGetFetchFunc(opts);
```
wilwade, line 182: I'm not sure we want to get into the fetch business. Although we are with the batch Parquet files (as we don't need to fetch the entire content).

There are just so few times that I can imagine that someone would want to call this function and not either have or need the activity content.

That said, it is validating the url and the data at the other end should be small.

Let's discuss with the team.

src/core/identifiers/identifiers.ts
wilwade: Maybe? Let's discuss with the team today.

src/core/announcements/validation.ts
```diff
@@ -0,0 +1,206 @@
...
+ * isAnnouncement() is a type check for Announcement
+ *
+ * @param obj - The object to check
+ * @returns True if the object is a Announcement, otherwise false
+ */
+const isAnnouncement = (obj: unknown): obj is Announcement => {
+  if (!isRecord(obj)) return false;
+  if (!isAnnouncementType(obj["announcementType"])) return false;
+
+  const validators = {
```
wilwade, line 146: We need to add this type so that TypeScript will throw an error if we add a new `AnnouncementType`
```suggestion
  const validators: Record<AnnouncementType, (obj: unknown) => boolean> = {
```

src/core/announcements/validation.ts
```diff
@@ -0,0 +1,206 @@
...
+  )
+    return false;
+
+  if (isBroadcastAnnouncement(obj) || isReplyAnnouncement(obj) || isProfileAnnouncement(obj)) {
+    const content = await fetch(obj["url"]).then((res) => res.text());
+    if (keccak256(content) != obj["contentHash"]) return false;
+
+    const activityContent = JSON.parse(content);
+    if (!isValidActivityContent(activityContent)) return false;
+  }
```
wilwade, line 203: Along the questions about fetch, I think this should be considered a separate piece (or at least `JSON.parse` and `isValidActivityContent`). Either that or perhaps `isValidAnnouncement` shouldn't be in core? This does feel like a composition method.

wilwade (dismissed):
Two discussions, a type fix, and maybe moving a method.

All the validations look great however. Really excited to see those. (Although I wouldn't be surprised if we have to write more soon)

src/core/announcements/validation.ts
rosalinekarr: removed

src/core/announcements/validation.ts
rosalinekarr: removed

src/createPublication.ts
```diff
@@ -1,14 +1,14 @@
...
-import { filterIterable, AsyncOrSyncIterable } from "./core/utilities";
-import { Publication } from "./core/contracts/publisher";
-import { getRandomString } from "./core/utilities/random";
 import {
   AnnouncementWithSignature,
   AnnouncementType,
   SignedAnnouncement,
   TypedAnnouncement,
 } from "./core/announcements";
+import { createFile, EmptyBatchError } from "./core/batch";
```
shannonwells, line 8: Especially for long PRs like this one, It would be nice to leave out formatting changes like this.

shannonwells (commented):
This is a lot, and there are files where nothing was done except move some imports from one line to another.  There's also some renaming. Is it possible to split this up into a rename + a validation PR, and drop the line-moving changes?

src/createPublication.ts
rosalinekarr: fixed

src/core/announcements/utilities.ts
```diff
@@ -0,0 +1,19 @@
+import { SignedAnnouncement } from "./crypto";
```
enddynayn, line 1: minor but this seems more like something that belongs a in a service.

rosalinekarr: any suggestion on where to put it instead?

src/generators/activityContentGenerators.ts
```diff
@@ -0,0 +1,81 @@
+import { sample, sampleText } from "@dsnp/test-generators";
+
+import {
```
enddynayn, line 3: what are the difference between a generator and a factory?

rosalinekarr: I think, within our repo at least, "factories" are just general factory functions which help the user create objects more easily and "generators" are functions specifically for creating fake objects for testing. "Factories" ask the user for all their data to go in the object except for what they can infer from types. "Generators" make up fake data to use.

rosalinekarr: That said, these terms are totally overloaded and messy, and I would be totally up for a rename.

src/core/announcements/utilities.ts
enddynayn: /service/....ts. or services.ts

src/generators/activityContentGenerators.ts
enddynayn: oohhh i see this was stuff from before. its ok.

enddynayn (approved):

src/core/announcements/utilities.ts
rosalinekarr: Updated

amparks100 (approved):

PR #96, opened 2021-07-20
Removed put from store interface

rosalinekarr:
No story for this, just something we've been needing to do that I knocked out in my spare time.

Problem
=======
We didn't know how we'd use the store adapter when we first wrote the interface for it, and now we have redundant put methods, `put()` and `putStream()`, where only one is really needed.

Solution
=======
Remove `put()` from the store interface

### DIFF
--- a/src/test/testStore.ts
+++ b/src/test/testStore.ts
@@ -7,11 +7,6 @@ export default class TestStore implements StoreInterface {
     this.store = {};
   }
 
-  async put(targetPath: string, content: Content): Promise<URL> {
-    this.store[targetPath] = content;
-    return new URL(`http://fakestore.org/${targetPath}`);
-  }
-
   async putStream(targetPath: string, callback: WriteStreamCallback): Promise<URL> {
     const buffers: Buffer[] = [];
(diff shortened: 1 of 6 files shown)


src/test/testStore.ts
```diff
@@ -7,27 +7,33 @@ export default class TestStore implements StoreInterface {
...
   async putStream(targetPath: string, callback: WriteStreamCallback): Promise<URL> {
     const buffers: Buffer[] = [];
     const writeStream = {
       write: (...args: unknown[]): boolean => {
         const [chunk, maybeCallback1, maybeCallback2] = args;
         if (typeof maybeCallback1 === "function") maybeCallback1(null);
         else if (typeof maybeCallback2 === "function") maybeCallback2(null);
-        buffers.push(chunk as Buffer);
+        if (typeof chunk === "string") {
+          buffers.push(Buffer.from(chunk));
```
enddynayn, line 18: Buffer.from, I don't think that would work on the browser.

enddynayn (approved):

wilwade (approved):

PR #98, opened 2021-07-22
Bugs/zero x prefixing 178965517

rosalinekarr:
Problem
=======
Ethers requires a 0x prefix on all hexadecimal values, and our code is inconsistent in hexadecimal prefixes.
[#178965517](https://www.pivotaltracker.com/story/show/178965517)

Solution
========
Update all hexadecimal values everywhere to start with a "0x"


enddynayn (approved):
fabulous 👍

src/core/utilities/hash.test.ts
```diff
@@ -0,0 +1,10 @@
+import { hash } from "./hash";
+
+describe("hash", () => {
+  describe("hash", () => {
```
acruikshank, line 4: Don't think you need this second describe.

acruikshank (approved):
Looks good.

src/core/utilities/hash.test.ts
rosalinekarr: Fixed

PR #105, opened 2021-07-23
DSNP Announcement Id -> DSNP Announcement Uri

wilwade:
Problem
=======
DSNP Announcement Id and User Id were out of spec naming and requirements

[178882099](https://www.pivotaltracker.com/story/show/178882099)

Solution
========
Update to match the v1.0 Spec

Double Checks:
---------------

Change summary:
---------------
* DSNP Announcement Id -> DSNP Announcement Uri
* DSNP User Id remove prefix
* DSNP User Uri is User Id with prefix
* DSNP User Id is allowed to trim leading zeros


src/core/identifiers/identifiers.ts
```diff
@@ -36,7 +43,18 @@ export type DSNPUserId = string;
...
+  return id.match(/^0x[0-9A-F]{1,16}$/i) !== null;
+};
+
+/**
+ * isDSNPUserUri validates a given object as a DSNPUserUri.
+ *
+ * @param uri - The object to validate
+ * @returns True of false depending on whether the string is a valid DSNPUserUri
+ */
+export const isDSNPUserUri = (uri: unknown): uri is DSNPUserUri => {
```
enddynayn, line 55: why no string instead of unknown?

enddynayn (approved):

src/core/identifiers/identifiers.ts
wilwade: This is a guard, so it takes an unknown and then validates it as a DSNPUserUri (aka string).

Guard standard is to start with an unknown.

src/core/identifiers/identifiers.test.ts
```diff
@@ -1,39 +1,48 @@
...
       "dsnp://0x0123456789abcdef/0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // Lowercase
     ];
 
-    const invalidDSNPAnnouncementIds = [
+    const invalidDSNPAnnouncementUris = [
       "dsnp://0123456789ABCDEF/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // No 0x on user
       "dsnp://0x0123456789abcdef/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on announcement
       "dsnp://0123456789abcdef/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on either
-      "dsnp://0xbadbadbad/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // Bad user id
+      "dsnp://0xbadwolf/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // Bad user id
```
rosalinekarr, line 14: [🐺](https://www.youtube.com/watch?v=9847_BEaLlM)

src/core/identifiers/identifiers.ts
```diff
@@ -36,7 +43,18 @@ export type DSNPUserId = string;
...
+  if (!isString(uri)) return false;
+  return uri.match(/^0x[0-9A-F]{1,16}$/i) !== null;
```
rosalinekarr, line 57: This is the same regex as `isDSNPUserId` above. I'm pretty sure this one is supposed to start with `dsnp://`.

CHANGELOG.md
```diff
@@ -5,17 +5,14 @@
...
 - Removed put method from the store interface in favor of putStream everywhere
 - Updated activityContent types to match spec
 - Updated activityContent generators to match new types
-- Added type check and validation functions to activityContent module
-- Added type check and validation functions to announcements module
 - Updated content porcelain to accept proper activityContent objects and validate them
 - Renamed activityPub -> activityContent
-- Renamed DSNPId to DSNPAnnouncementId to avoid confusion with DSNPUserId
-- Renamed uri -> url everywhere
+- Renamed DSNPId to DSNPAnnouncementUri to avoid confusion with DSNPUserId
```
rosalinekarr, line 14: Given that we all caps `DSNP`, shouldn't we caps `Uri` as well? So this should be `DSNPAnnouncementURI` everywhere?

src/content.test.ts
```diff
@@ -20,7 +20,7 @@ describe("content", () => {
         store = new TestStore();
 
         setConfig({
-          currentFromId: "dsnp://0x0123456789ABCDEF",
+          currentFromId: "0x0123456789ABCDEF",
```
rosalinekarr, line 23: Since the user is going to have to set `currentFromId` manually, why don't we just change it to `currentFromURI` and store a full DSNPUserURI instead so the user never has to convert between URIs and Ids?

src/core/contracts/registry.test.ts
```diff
@@ -51,7 +51,7 @@ describe("registry", () => {
     });
 
     it("Returns the correct id", async () => {
-      expect(givenId).toEqual("dsnp://0x00000000000003e8");
+      expect(givenId).toEqual("0x03e8");
```
rosalinekarr, line 54: Wait, I thought the plan was to only ever return easier-to-use URIs for DSNPIds to the end user, not unlabelled hex strings. This string is ultimately returned to the user.

Am I remembering incorrectly or was there some change of plans?

src/createPublication.test.ts
```diff
@@ -63,28 +63,28 @@ describe("createPublications", () => {
   let store: TestStore;
   const messages = [
     createBroadcast(
-      "dsnp://0x0123456789ABCDEF",
+      "0x0123456789ABCDEF",
```
rosalinekarr, line 66: Same question, but on the input side here. I thought we wanted users to only ever input DSNPUserURIs, not DSNPUserIds, since URIs are harder to mistake for other hex strings.

rosalinekarr (dismissed):

src/core/identifiers/identifiers.ts
rosalinekarr: fixed

src/content.test.ts
rosalinekarr: fixed

src/core/contracts/registry.test.ts
rosalinekarr: fixed

src/createPublication.test.ts
rosalinekarr: fixed

CHANGELOG.md
rosalinekarr: fixed

amparks100 (approved):

shannonwells (approved):
little too late, but checks out

PR #110, opened 2021-07-28
check for pastLogs.length before continuing in subscribeToBatchPublications

shannonwells:
Problem
=======
In `subscribeToBatchPublications`, we needed to check the length of pastLogs before trying to look at the blockNumber of one of its elements.
[Fixes #179033826](https://www.pivotaltracker.com/story/show/179033826)

Solution
========
Do the length check
Add a regression test that shows the red/green behavior

Double Checks:
---------------

Steps to Verify:
----------------
1. The test passes
1. For extra credit you can build sdk and make example-client use it, then launch a hardhat localnet with no logs or anything, and login to the example client.  You should not see the error message as shown in the Tracker story.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 ## Unreleased
 ### Changes
+- Fix length check bug in `subscribeToBatchPublications`
 - Updated announcement types to include createdAt type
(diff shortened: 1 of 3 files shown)


enddynayn (approved):

aneyzberg (approved):
lgtm

PR #116, opened 2021-08-03
Not all ethereum nodes will return a message with why something reverted

wilwade:
Problem
=======
OpenEthereum can be configured to not return the revert error message which tripped up the resolveRegistration code testing for it.

Solution
========
Removed the check for the error message

with @amparks100 

Double Checks:
---------------

Change summary:
---------------
* Updated registry.resolveRegistration to support not found cases for nodes that do not return a failure reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+### Changes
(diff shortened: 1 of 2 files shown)


enddynayn (approved):
Wow. Just wow. 👍

amparks100 (approved):

PR #117, opened 2021-08-03
No longer using destructuring

pfrank13:
Now not using destructuring because of, what I can only assume, is
"this" style issues. Basically a private member _writableState will be
undefined. Using the object directly fixes that issue

Problem
=======
using destructuring causes FileStreams to have undefined members.
[Pivotal issue](https://www.pivotaltracker.com/story/show/179110784)

Solution
========
No longer use destructuring and call the WriteStream methods .write and .end directly

Double Checks:
---------------

Change summary:
---------------
No longer using destructuring when working with the WriteStream in the putStream callback flow

Steps to Verify:
----------------
A bit tough as I needed to consume the library in another app to make sure it worked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 ### Changes
 - Updated registry.resolveRegistration to support not found cases for nodes that do not return a failure reason.
 
+## [2.0.1] - TBD
+### Fixed
+- no longer using destructuring for methods that result in a `store.putStream` call. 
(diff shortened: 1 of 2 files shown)


enddynayn (approved):

shannonwells (approved):
LGTM, It's a safe change regardless as the WriteStream requires both functions & will error beforehand if they don't exist on the callback.

PR #120, opened 2021-08-04
stuff

enddynayn:
Problem
=======
problem statement, including
[link to Pivotal Tracker #12345678](https://www.pivotaltracker.com/story/show/12345678)

Solution
========
What I/we did to solve this problem

with @pairperson1

Double Checks:
---------------
- [] Any new modules need to be exported?
- [] Are new methods in the right module?
- [] Are new methods/modules in core or not (i.e. porcelain or plumbing)?
- [] Do you have good documentation on exported methods?

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing


src/core/contracts/subscription.ts
```diff
@@ -19,6 +20,20 @@ export interface BatchPublicationCallbackArgs {
   fileHash: HexString;
 }
```
rosalinekarr, line 21: Can we delete this now?

src/core/contracts/identity.test.ts
```diff
@@ -656,19 +664,22 @@ describe("identity", () => {
...
           blockNumber: 332,
+          transactionHash: expect.any(String),
         },
         {
           name: "DSNPRemoveDelegate",
           identityAddress: "0x8464135c8F25Da09e49BC8782676a84730C318bC",
           delegate: "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
           blockNumber: 333,
           endBlock: 400,
+          transactionHash: expect.any(String),
```
rosalinekarr, line 682: not sure if it makes sense here, but you could do a regex match here to make sure it's a hash

```
transactionHash: expect.match(/^0x[0-9a-f]{64}/i)
```

rosalinekarr (approved):
A couple non-blocking comments, but otherwise looks good

wilwade (changes requested):
All the code looks good! Don't forget the changelog however!

wilwade (approved):
👍

src/core/contracts/identity.test.ts
enddynayn: I like it but not sure if it adds much value for this test.

PR #122, opened 2021-08-04
v2.0.1

wilwade:
### Changes
- Updated registry.resolveRegistration to support not found cases for nodes that do not return a failure reason.
- Updated the Activity Content Published field validation regex to support fractional seconds
- Updated @dsnp/contracts to v1.0.1
- Updated minor dependencies

### Fixed
- no longer using destructuring for methods that result in a `store.putStream` call.
- sdk.createRegistration was incorrectly returning a DSNP User Id instead of a DSNP User URI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,17 @@
 
 ## Unreleased
 ### Changes
+
+## [2.0.1] - 2021-08-04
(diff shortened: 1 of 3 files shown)


aneyzberg (approved):
lgtm

PR #123, opened 2021-08-06
dsnpBatchFilter does not require any provider or signer

wilwade:
Problem
=======
The announcements API has a single method to get the contract for batch and subscribe operations. This method requires a signer to connect. This is necessary for the batch operation. The subscription (specifically the dsnpBatchFilter) is read only, so should be capable of running with just a provider.
[#178901784](https://www.pivotaltracker.com/story/show/178901784)

Solution
========
Used the interface instead of the contract to get the needed topic data.

Double Checks:
---------------

Change summary:
---------------
* Switched dsnpBatchFilter to use the interface instead of the contract to create the filter.

### DIFF
--- a/src/core/contracts/publisher.ts
+++ b/src/core/contracts/publisher.ts
@@ -3,6 +3,7 @@ import { ConfigOpts, requireGetProvider, requireGetSigner } from "../config";
 import { HexString } from "../../types/Strings";
 import { Publisher, Publisher__factory } from "../../types/typechain";
 import { getContractAddress } from "./contract";
+import { AnnouncementType } from "../announcements";
 
 const CONTRACT_NAME = "Publisher";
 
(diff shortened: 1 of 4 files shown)


src/core/contracts/publisher.ts
```diff
@@ -33,17 +33,14 @@ export const publish = async (publications: Publication[]): Promise<ContractTran
...
- * @throws {@link MissingProviderConfigError}
- * Thrown if the provider is not configured.
- * @throws {@link MissingContractAddressError}
- * Thrown if the batch contract address cannot be found.
  * @returns DSNPBatch event filter
  */
-export const dsnpBatchFilter = async (): Promise<EventFilter> => {
-  const contract = await getPublisherContract();
-  return contract.filters.DSNPBatchPublication();
+export const dsnpBatchFilter = (): EventFilter => {
```
enddynayn, line 38: can we we also pass in as an optional argument the annoucementTypeTopic

enddynayn: also maybe combine it with

```
const createFilter = (batchFilter: ethers.EventFilter, filterOptions: BatchFilterOptions) => {
  const topics = batchFilter.topics ? batchFilter.topics : [];
  const announcementTypeTopic = filterOptions?.announcementType
    ? "0x" + filterOptions.announcementType.toString(16).padStart(64, "0")
    : null;
  if (announcementTypeTopic) {
    topics.push(announcementTypeTopic);
  }

  const finalFilter: ethers.providers.EventType = {
    topics: topics,
  };
  return finalFilter;
};
```

enddynayn (approved):

enddynayn:
👍 beautiful

PR #124, opened 2021-08-06
refac(batch-subscription): extract common code 

enddynayn:
Extract code that gets past logs and
subscribes to new log events into
#subscribeToEvent function.

This allows the ability subscribe
and fetch past logs to used for
other event subscriptions.

### DIFF
--- a/src/core/contracts/subscription.test.ts
+++ b/src/core/contracts/subscription.test.ts
@@ -3,8 +3,9 @@ import { publish, Publication, dsnpBatchFilter } from "./publisher";
 import { subscribeToBatchPublications, BatchPublicationLogData, BatchFilterOptions } from "./subscription";
 import { setupSnapshot } from "../../test/hardhatRPC";
 import { setupConfig } from "../../test/sdkTestConfig";
-import { checkNumberOfFunctionCalls } from "../../test/utilities";
+import { checkNumberOfFunctionCalls, mineBlocks } from "../../test/utilities";
 import { hash } from "../utilities";
+import { ethers } from "ethers";
 
 describe("subscription", () => {
   setupSnapshot();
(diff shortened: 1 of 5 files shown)


src/core/contracts/subscription.test.ts
```diff
@@ -64,21 +65,66 @@ describe("subscription", () => {
...
+          expect(provider.listeners(filter).length).toEqual(0);
+        });
+      });
 
-        const removeListener = await subscribeToBatchPublications(mock, filterOptions);
-        expect(mock).not.toHaveBeenCalled();
-        await removeListener();
-        const filter = await dsnpBatchFilter();
-        expect(provider.listeners(filter).length).toEqual(0);
+      describe("and fromBlock is from the future", () => {
```
enddynayn, line 88: update test

src/core/contracts/utilities.ts
```diff
@@ -77,3 +78,46 @@ export interface LogEventData {
   transactionHash: HexString;
   blockNumber: number;
 }
+
+export const subscribeToEvent = async (
+  provider: ethers.providers.Provider,
```
enddynayn, line 83: dont need to pass this in as argument

wilwade: We could do a getRequireProvider, but that would require potentially adding ConfigOps. I like passing it in and requiring it at the higher level, but we do mostly follow the ConfigOpts pattern.

src/core/contracts/subscription.ts
```diff
@@ -56,67 +55,16 @@ export const subscribeToBatchPublications = async (
...
-    topics: topics,
-  };
-  return finalFilter;
-};
+  const batchFilter: ethers.EventFilter = await dsnpBatchFilter(filter?.announcementType);
 
-const getPastLogs = async (provider: ethers.providers.Provider, filter: Filter): Promise<BatchPublicationLogData[]> => {
-  const logs = await provider.getLogs(filter);
-  return decodeLogsForBatchPublication(logs);
+  return subscribeToEvent(provider, batchFilter, doReceiveEvent, filter?.fromBlock);
```
enddynayn, line 67: dont pass provider argument. ill leave more abstraction for another time.

PR #127, opened 2021-08-09
Add flexibility to allowing input of non-0x prefixed hex strings

wilwade:
Problem
=======
Some things (node buffer) don't 0x prefix and need to match spec clarification https://github.com/LibertyDSNP/spec/pull/113

Solution
========
Allow non-0x prefixed strings to be handled by convertDSNPUserIdOrURIToBigNumber

Double Checks:
---------------

Change summary:
---------------
* Allow non-0x prefixed strings to be handled byconvertDSNPUserIdOrURIToBigNumber
* Only lowercase allowed
* Update to match DSNP spec 0.9.2 https://github.com/LibertyDSNP/spec/pull/113


src/core/identifiers/identifiers.ts
```diff
@@ -1,8 +1,9 @@
 import { BigNumber } from "ethers";
-
 import { isString } from "../utilities/validation";
 import { HexString } from "../../types/Strings";
 
+const JUST_HEX_REGEX = /dsnp:\/\/0x0*|0x0*/gi;
```
rosalinekarr, line 5: how about we rename this to `HEX_PREFIX_REGEX` or something else more descriptive?

rosalinekarr: also, it might be a good idea to add a `^` to the beginning to make sure it's at the start of the string

src/core/identifiers/identifiers.ts
```diff
@@ -18,7 +19,7 @@ export type DSNPAnnouncementURI = string;
  */
 export const isDSNPAnnouncementURI = (id: unknown): id is DSNPAnnouncementURI => {
   if (!isString(id)) return false;
-  return id.match(/^dsnp:\/\/0x[0-9A-F]{1,16}\/0x[0-9A-F]{64}$/i) !== null;
+  return id.match(/^dsnp:\/\/0x[1-9a-f][0-9a-f]{0,15}\/0x[1-9a-f][0-9a-f]{0,63}$/) !== null;
```
rosalinekarr, line 22: we did we drop case-insensitivity here? are we requiring that hex be lowercase too?

src/core/identifiers/identifiers.ts
```diff
@@ -43,7 +44,7 @@ export type DSNPUserURI = string;
  */
 export const isDSNPUserId = (id: unknown): id is DSNPUserId => {
   if (!isString(id)) return false;
-  return id.match(/^0x[0-9A-F]{1,16}$/i) !== null;
+  return id.match(/^0x[1-9a-f][0-9a-f]{0,15}$/) !== null;
```
rosalinekarr, line 47: same thing here

src/core/identifiers/identifiers.ts
```diff
@@ -54,7 +55,7 @@ export const isDSNPUserId = (id: unknown): id is DSNPUserId => {
  */
 export const isDSNPUserURI = (uri: unknown): uri is DSNPUserURI => {
   if (!isString(uri)) return false;
-  return uri.match(/^dsnp:\/\/0x[0-9A-F]{1,16}$/i) !== null;
+  return uri.match(/^dsnp:\/\/0x[1-9a-f][0-9a-f]{0,15}$/) !== null;
```
rosalinekarr, line 58: and here

src/core/identifiers/identifiers.ts
```diff
@@ -75,8 +76,8 @@ export const convertBigNumberToDSNPUserId = (num: BigNumber): DSNPUserId => num.
  */
 export const convertDSNPUserIdOrURIToBigNumber = (userIdOrUri: DSNPUserId | DSNPUserURI): BigNumber => {
   // Support DSNP User URI
-  const hex = userIdOrUri.replace("dsnp://", "");
-  return BigNumber.from(hex);
+  const hex = userIdOrUri.replace(/dsnp:\/\/|0x/gi, "");
```
rosalinekarr, line 79: why do we remove the `0x` only to add it back in the next line?

rosalinekarr: also, why is this regex global? there should only ever be one DSNP prefix

rosalinekarr: on and the `^` thing too

src/core/identifiers/identifiers.ts
```diff
@@ -85,8 +86,9 @@ export const convertDSNPUserIdOrURIToBigNumber = (userIdOrUri: DSNPUserId | DSNP
  * @param userURI - The DSNPUserURI to convert
  * @returns The DSNPUserId of the user
  */
-export const convertDSNPUserURIToDSNPUserId = (userURI: DSNPUserURI): DSNPUserId =>
-  userURI.replace(/dsnp:\/\/0x0+/, "0x0");
+export const convertDSNPUserURIToDSNPUserId = (userURI: DSNPUserURI): DSNPUserId => {
+  return serializeToHex(userURI);
+};
```
rosalinekarr, line 91: if this function is just an alias for `serializeToHex`, that makes me think we should delete one and just use the other

src/core/identifiers/identifiers.ts
```diff
@@ -133,3 +129,20 @@ export const parseDSNPAnnouncementURI = (
   const [userId, contentHash] = announcementUri.replace("dsnp://", "").split("/");
   return { userId, contentHash };
 };
+
+export const serializeToHex = (
```
rosalinekarr, line 133: I realize this is partially tested through all the other functions in this file, but can we test `serializeToHex()` explicitly?

src/core/identifiers/identifiers.test.ts
```diff
@@ -1,20 +1,56 @@
...
     const invalidDSNPAnnouncementURIs = [
-      "dsnp://0123456789ABCDEF/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // No 0x on user
-      "dsnp://0x0123456789abcdef/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on announcement
+      "dsnp://0xa123456789ABCDEF/0xa123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // Uppercase
+      "dsnp://0x0000456789abcdef/0x0000456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // Leading Zeros
+      "dsnp://0123456789abcdef/0xa123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on user
+      "dsnp://0xa123456789abcdef/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on announcement
       "dsnp://0123456789abcdef/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef", // No 0x on either
-      "dsnp://0xbadwolf/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // Bad user URI
-      "dsnp://0x0123456789ABCDEF/0xbadbadbad", // Bad message id
```
rosalinekarr, line 15: why did we drop the bad message id test?

src/core/identifiers/identifiers.test.ts
```diff
@@ -1,20 +1,56 @@
...
+    for (const id of invalidDSNPUserURIs) {
+      it(`returns false for "${id}"`, () => {
+        expect(isDSNPUserURI(id)).toEqual(false);
+      });
+    }
+  });
+
   describe("isDSNPAnnouncementURI", () => {
     const validDSNPAnnouncementURIs = [
-      "dsnp://0x0123456789ABCDEF/0x0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF", // Uppercase
```
rosalinekarr, line 6: why did we drop the uppercase test?

rosalinekarr (changes requested):
a few questions and comments, but otherwise looks good

src/core/identifiers/identifiers.ts
wilwade: Global drop 👍 
^ 👍 
Remove to add? Because the regex is using it to drop leading zeros and it is a more complex regex to only drop the leading zeros.

src/core/identifiers/identifiers.ts
wilwade: There are a few of these. I wanted to keep them for backward compatibility and also for better typing and naming.

src/core/identifiers/identifiers.ts
wilwade: Ah forgot about docs on it as well. It was originally not exported. I'll fix. Or I think I will move to a different module and add tests and docs.

src/core/identifiers/identifiers.test.ts
wilwade: `0xbadbadbad` is valid now.

src/core/identifiers/identifiers.test.ts
wilwade: The spec says these are invalid. It always specified lowercase and I think we just missed it first time around.

src/core/identifiers/identifiers.test.ts
rosalinekarr: could we add one with some non-hex chars then? maybe `dsnp://0x1234567890/0xNotHexadecimal`?

src/core/announcements/errors.ts
```diff
@@ -21,3 +21,17 @@ export class InvalidAnnouncementTypeError extends AnnouncementError {
     this.announcementType = announcementType;
   }
 }
+
+/**
+ * InvalidHexadecimalSerialization indicates an attempt to coerce a string to hexadecimal serialization failed.
+ */
+export class InvalidHexadecimalSerialization extends DSNPError {
```
rosalinekarr, line 28: we might want to move this to `sdk/core/errors` since it's used in more than just the announcement module

rosalinekarr (approved):
LGTM, only one non-blocking item

src/core/identifiers/identifiers.ts
wilwade: I want the specific types however.

src/core/identifiers/identifiers.test.ts
wilwade: 👍 new Line 52

CHANGELOG.md
```diff
@@ -19,19 +31,13 @@
...
 - Updated subscribeToBatchPublications to not need a signer
 - sdk.core.contracts.publisher.dsnpBatchFilter is no longer async
 - Updated Ethers from 5.3.0 -> 5.4.4
 
-### Fixed
-- Fixed bug resulting in incorrect output from sdk.core.contracts.publisher.dsnpBatchFilter when passed Tombstone
-
 ## [2.0.1] - 2021-08-04
 ### Changes
 - Reworked ActivityContent validations to log or throw informative errors when an ActvityContent is invalid. Fixed a bug in DURATION_REGEX.
```
shannonwells, line 43: This is the line that needs to be removed :P

src/core/announcements/serialization.ts
```diff
@@ -8,12 +11,56 @@ import { Announcement } from "./factories";
...
+export const serializeToHex = (
+  value: BigNumber | HexString | DSNPUserId | DSNPUserURI | number | bigint | unknown
+): HexString => {
+  if (typeof value === "number" || typeof value === "bigint") {
+    return prefixAndStrip(value.toString(16));
+  }
+  if (BigNumber.isBigNumber(value)) {
+    return prefixAndStrip(value.toHexString());
+  }
+  if (String(value) === "") {
```
shannonwells, line 37: FWIW I just checked this bit of code with a value created from the Long library, which is what's used by parquet, and it works.

src/core/announcements/validation.test.ts
```diff
@@ -51,29 +52,47 @@ describe("validation", () => {
...
         const followeeTransaction = await register(followeeIdentityContract.address, "George Bluth");
-        followeeId = await getURIFromRegisterTransaction(followeeTransaction);
+        followeeId = serializeToHex(await getURIFromRegisterTransaction(followeeTransaction));
       });
 
       afterAll(async () => {
         await revertHardhat(provider);
       });
 
       it("returns true for valid graph change announcements", async () => {
```
shannonwells, line 62: does there need to be one test that uses the output of createFollowGraphChange?

src/core/errors/errors.ts
```diff
@@ -9,3 +9,17 @@ export class DSNPError extends Error {
...
+
+/**
+ * InvalidHexadecimalSerialization indicates an attempt to coerce a string to hexadecimal serialization failed.
+ */
+export class InvalidHexadecimalSerialization extends DSNPError {
+  outcomeValue: string;
+  originalValue: unknown;
+
+  constructor(originalValue: unknown, outcomeValue: string) {
+    super(`Invalid Hexadecimal Serialization: ${outcomeValue}`);
```
shannonwells, line 21: What's done with `originalValue`? Looks like it's unused.

src/core/utilities/hash.ts
```diff
@@ -8,7 +8,7 @@ import { HexString } from "../../types/Strings";
  * @param content - The string content to hash
  * @returns A 0x prefixed keccak256 hash
  */
-export const hash = (content: string): HexString => `0x${keccak256(content)}`;
+export const hash = (content: Message): HexString => `0x${keccak256(content)}`;
```
shannonwells, line 11: I had to look up what a Message was but ❤️

shannonwells (commented):
LGTM okay by me, non-blocking feedback

shannonwells (approved):
👍

src/core/errors/errors.ts
wilwade: It's mostly to pass it through for if people want to see what the non-stringified version is

src/core/announcements/validation.test.ts
wilwade: Yes, but that should be over in the factories test. I'll add one over there.

src/core/announcements/serialization.test.ts
```diff
@@ -1,15 +1,95 @@
...
+import { BigNumber } from "ethers";
+import { InvalidHexadecimalSerialization } from "../errors";
 
 describe("serialization", () => {
   describe("#serialize", () => {
-    it("returns the correct serialization for a given announcement", async () => {
-      const announcement = createBroadcast("1", "https://dsnp.org", "0x12345");
-      const serializeAnnouncement = await serialize(announcement);
+    it("returns the correct serialization for a given announcement", () => {
+      const announcement = createBroadcast("dsnp://0x1", "https://dsnp.org/", "0x12345");
```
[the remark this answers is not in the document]
wilwade: Chance.

PR #130, opened 2021-08-10
Features/creating tombstones 179117207

rosalinekarr:
Problem
=======
We want to be able to create Tombstone announcements with the SDK now that they're part of the spec.
[#179117207](https://www.pivotaltracker.com/story/show/179117207)

Solution
========
Add the ability to create tombstones.


shannonwells (approved):
Looks good, small non-blocking request

PR #132, opened 2021-08-12
Bugs/better errors #179106793

shannonwells:
Problem
=======
We need more informative errors.
[Fixes #179106793](https://www.pivotaltracker.com/story/show/179106793)

Solution
========
Since the only entry points to the validations threw an error if the ActivityContent(Note|Profile) was invalid, almost everything now throws an error if it doesn't validate, however, sometimes we only need one item of a list to be valid, in particular, attachments.  Attachment validation has been abstracted.

1. Anything that throws an error when something is not valid has been renamed to `requireFoo`
1. Functions that do simple type checking without validating contents should end with `Type`
1. If an attachment has multiple URLs, only one need be valid.  The following examples would be valid:
    1. One url has only unsupported hash algorithms - MD5 + secp256k1, but the other has keccak256 which is supported.
    1. One url href  is `ftp` but the other is `https` 
1. If a Note or Profile has multiple attachments, only one need be valid. The following examples would be valid:
    1. Two image attachments: one has only a url pointing to a HEIC (unsupported) and the other points to a JPEG (supported)
    2. Two image attachments:  one points to an FTP URL (unsupported) and the other is an HTTPS URL (supported).
1. A Note or Profile can have an empty attachments array and be valid, but if it doesn't, and _none_ are valid, we rerun our validations without catching errors and throw the first one.  Example:   An Image attachment for a Note has only an FTP url, the validation will throw an `InvalidActivityContentError` saying `ActivityContentLink href is invalid`.
1. The new function that is used above,  `requireGetSupportedContentAttachments` is also exported for use by SDK developers to assist in debugging or simply to get a list of all the attachments valid for DSNP.
1. In order to do the above, on the first pass, though attachment validations throw errors, they are caught and logged to console, for ease of debugging.

Error/return value behavior isn't 100% consistent; there's too much going on and needs are different depending on the object's depth.

with @rosalinekarr 

Double Checks:
---------------
- [x] Any new modules need to be exported?  NO
- [x] N/AAre new methods/modules in core or not (i.e. porcelain or plumbing)?

Change summary:
---------------
* Rework and rename validators to throw or log informative errors when an `ActivityContent` is invalid
* Redo tests to reflect changes, add more to improve test coverage
* Fix a bug in `DURATION_REGEX` 

Steps to Verify:
----------------
* Please help by reviewing for consistent naming.
* All tests should pass -- see test coverage
* There are tests for as many potential error cases (modulo being in different attachments), so if you think of one I've left out please comment.

### DIFF
--- a/src/core/activityContent/validation.test.ts
+++ b/src/core/activityContent/validation.test.ts
@@ -1,13 +1,398 @@
-import { ActivityContentNote, ActivityContentProfile } from "./factories";
+import { ActivityContentAttachment, ActivityContentNote, ActivityContentProfile } from "./factories";
 import {
   requireActivityContentNoteType,
-  requireIsActivityContentProfile,
+  requireIsActivityContentProfileType,
   requireValidActivityContentNote,
   requireValidActivityContentProfile,
+  requireGetSupportedContentAttachments,
 } from "./validation";
 
 describe("activity content validations", () => {
-  describe("isActivityContentNote", () => {
+  describe("requireGetSupportedContentAttachments", () => {
+    describe("when there is one valid attachment", () => {
+      const testCases: Record<string, Array<ActivityContentAttachment>> = {
+        "Image attachment with multiple urls and only one is valid": [
+          {
+            type: "Image",
+            url: [
+              {
+                type: "Link",
+                href: "https://upload.wikimedia.org/wikipedia/commons/a/ae/Mccourt.jpg",
+                mediaType: "image/jpg",
+                hash: [
+                  {
+                    algorithm: "keccak256",
+                    value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                  },
+                ],
+              },
+              {
+                type: "Link",
+                href: "https://upload.wikimedia.org/wikipedia/commons/a/ae/Mccourt.jpg",
+                mediaType: "image/foo",
+                hash: [
+                  {
+                    algorithm: "keccak256",
+                    value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+        "Video attachment with multiple urls but only one has a valid hash algorithm": [
+          {
+            type: "Video",
+            name: "My video",
+            duration: "00:00:47",
+            url: [
+              {
+                type: "Link",
+                href: "https://upload.wikimedia.org/wikipedia/commons/c/c0/Big_Buck_Bunny_4K.webm",
+                mediaType: "video/webm",
+                hash: [
+                  {
+                    algorithm: "keccak256",
+                    value: "0xf841950dfcedc968dbd63132da844b9f28faea3dbfd4cf326b3831b419a20e9a",
+                  },
+                  {
+                    algorithm: "secp256k1",
+                    value: "0xf841950dfcedc968dbd63132da844b9f28faea3dbfd4cf326b3831b419a29999",
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+        "with a link attachment": [
+          {
+            type: "Link",
+            href: "https://dsnp.org",
+          },
+        ],
+        "with an audio attachment": [
+          {
+            type: "Audio",
+            url: [
+              {
+                type: "Link",
+                href: "https://upload.wikimedia.org/wikipedia/commons/d/d9/Wilhelm_Scream.ogg",
+                mediaType: "audio/ogg",
+                hash: [
+                  {
+                    algorithm: "keccak256",
+                    value: "0x3b33df3d163e86514e9041ac97e3d920a75bbafa8d9c1489e631897874b762cc",
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      };
+      for (const tc in testCases) {
+        it(`returns the attachment ${tc}`, () => {
+          expect(requireGetSupportedContentAttachments(testCases[tc])).toEqual(testCases[tc]);
+        });
+      }
+    });
+    describe("when there is one invalid attachment", () => {
+      [
+        {
+          name: "an audio attachment with a bad duration value",
+          expErr: "ActivityContentAudio duration is invalid",
+          attachment: [
+            {
+              name: "I have a name",
+              duration: "53 minutes",
+              type: "Audio",
+              url: [
+                {
+                  type: "Link",
+                  href: "https://upload.wikimedia.org/wikipedia/commons/d/d9/Wilhelm_Scream.ogg",
+                  mediaType: "audio/ogg",
+                  hash: [
+                    {
+                      algorithm: "keccak256",
+                      value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                    },
+                  ],
+                },
+              ],
+            },
+          ],
+        },
+        {
+          name: "an image attachment with a bad name value",
+          expErr: "ActivityContentImage name is not a string",
+          attachment: [
+            {
+              type: "Image",
+              name: 1234,
+              url: [
+                {
+                  type: "Link",
+                  href: "https://upload.wikimedia.org/wikipedia/commons/a/ae/Mccourt.jpg",
+                  mediaType: "image/jpg",
+                  hash: [
+                    {
+                      algorithm: "keccak256",
+                      value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                    },
+                  ],
+                },
+              ],
+            },
+          ],
+        },
+        {
+          name: "a video attachment with an unsupported mediaType",
+          expErr: "ActivityContentVideo has no supported media types",
+          attachment: [
+            {
+              type: "Video",
+              url: [
+                {
+                  type: "Link",
+                  href: "https://upload.wikimedia.org/wikipedia/commons/c/c0/Big_Buck_Bunny_4K.webm",
+                  mediaType: "video/avi",
+                  hash: [
+                    {
+                      algorithm: "keccak256",
+                      value: "0xf841950dfcedc968dbd63132da844b9f28faea3dbfd4cf326b3831b419a20e9a",
+                    },
+                  ],
+                },
+              ],
+            },
+          ],
+        },
+        {
+          name: "a link attachment with an invalid protocol",
+          expErr: "ActivityContentLink href is invalid",
+          attachment: [
+            {
+              type: "Link",
+              href: "ftp://dsnp.org",
+            },
+          ],
+        },
+        {
+          name: "audio attachment with an unsupported algorithm",
+          expErr: "ActivityContent hash algorithms must contain at least one of: keccak256",
+          attachment: [
+            {
+              type: "Audio",
+              url: [
+                {
+                  type: "Link",
+                  href: "https://upload.wikimedia.org/wikipedia/commons/d/d9/Wilhelm_Scream.ogg",
+                  mediaType: "audio/ogg",
+                  hash: [
+                    {
+                      algorithm: "secp256k1",
+                      value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                    },
+                  ],
+                },
+              ],
+            },
+          ],
+        },
+        {
+          name: "audio attachment with undefined hash",
+          expErr: "ActivityContentAudioLink hash is invalid",
+          attachment: [
+            {
+              type: "Audio",
+              url: [
+                {
+                  type: "Link",
+                  href: "https://upload.wikimedia.org/wikipedia/commons/d/d9/Wilhelm_Scream.ogg",
+                  mediaType: "audio/ogg",
+                },
+              ],
+            },
+          ],
+        },
+        {
+          name: "a link attachment with no type",
+          expErr: "ActivityContentAttachment type is invalid",
+          attachment: [
+            {
+              href: "https://dsnp.org",
+            },
+          ],
+        },
+      ].forEach((testCase) => {
+        it(`${testCase.name} throws error`, () => {
+          expect(() => requireGetSupportedContentAttachments(testCase.attachment)).toThrowError(testCase.expErr);
+        });
+      });
+    });
+    describe("when there are multiple attachments", () => {
+      it("with a single invalid attachment url, throws expected error", () => {
+        const missingMediaType = [
+          {
+            type: "Image",
+            url: [
+              {
+                type: "Link",
+                href: "https://upload.wikimedia.org/wikipedia/commons/a/ae/Mccourt.jpg",
+                hash: [
+                  {
+                    algorithm: "keccak256",
+                    value: "0x90b3b09658ec527d679c2de983b5720f6e12670724f7e227e5c360a3510b4cb5",
+                  },
+                ],
+              },
+            ],
+          },
+        ];
+        expect(() => requireGetSupportedContentAttachments(missingMediaType)).toThrowError(
+          "DSNPError: Invalid ActivityContent: ActivityContentImageLink mediaType is not a string"
+        );
+      });
+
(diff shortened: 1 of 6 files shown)


src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
-const isActivityContentMention = (obj: unknown): obj is ActivityContentMention => {
-  if (!isRecord(obj)) return false;
-  if (obj["type"] !== "Mention") return false;
-  if (!isDSNPUserId(obj["id"])) return false;
-
-  if (obj["name"] && !isString(obj["name"])) return false;
+const getActivityContentAttachment = (obj: unknown): ActivityContentAttachment | undefined => {
+  try {
+    return requireActivityContentAttachment(obj);
+  } catch (e) {
```
rosalinekarr, line 99: Can we change this to specifically catch `InvalidActivityContentError` so it doesn't catch real bugs in case we screw something up in the future?

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
+    console.error(e.toString());
```
rosalinekarr, line 100: Also, I don't think we want to log here since it could be expected behavior that a Note has no attachments and this function returns nothing.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
+  const duration = requireIsString(obj["duration"], `${contentType} duration`);
+  if (!isValidDurationField(duration)) throw new InvalidActivityContentError(`${contentType} duration is invalid`);
+};
 
-  return true;
+/**
+ * requireActivityContentAttachment validates an unknown object and returns it
+ * as an ActivityContentAttachment type if valid, and throws an error if not.
+ *
+ * @param obj - the object to convert
```
rosalinekarr, line 120: Can you document the `@throws` too?

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
+    case "Image":
+      return requireActivityContentImage(obj);
+    case "Audio":
+      return requireActivityContentAudio(obj);
+    case "Link":
+      return requireIsActivityContentLink(obj);
+    case "Video":
+      return requireActivityContentVideo(obj);
+    default:
+      throw new InvalidActivityContentError("unknown activity content type: " + obj["type"]);
```
rosalinekarr, line 135:
```suggestion
      throw new InvalidActivityContentError("unknown activity content attachment type: " + obj["type"]);
```

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
- * ActivityContentProfile objects. Note that this function only checks that the
- * given object meets the type definition of an ActivityContentProfile. It
- * does not** perform any logic validations, such as checking the format of
- * string fields or checking that the icon field includes at least one supported
- * image format.
- *
- * @param obj - The object to check
- * @returns A boolean indicating whether or not the object is an ActivityContentProfile
- */
-export const isActivityContentProfile = (obj: unknown): obj is ActivityContentProfile => {
```
rosalinekarr, line 208: Can we keep the non-error throwing versions of `isActivityContentNote` and `isActivityContentProfile` specifically? The have some value as type checks when we don't necessarily want to validate.

rosalinekarr (changes requested):

src/core/activityContent/validation.ts
shannonwells: Can you clarify?  If a note has no attachments then there won't be a console error.

rosalinekarr: Never mind, I just realized this function isn't exported. I was originally thinking users might use it.

src/core/activityContent/validation.test.ts
```diff
@@ -175,17 +293,82 @@ describe("activity content validations", () => {
...
-        "@context": "https://www.w3.org/ns/activitystreams",
-        type: "Note",
-        content: "What an adventure!",
-        mediaType: "text/plain",
-        attachment: [
-          {
-            type: "Video",
-            url: {
+        ];
+        expect(requireGetSupportedContentAttachments(multipleUrlsAttachment)).toStrictEqual(multipleUrlsAttachment);
```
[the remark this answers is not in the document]
shannonwells: No this one is a single attachment with multiple URLs, and only one is valid. But we only filter attachments, not their URLs.

src/core/activityContent/validation.test.ts
```diff
@@ -175,17 +293,82 @@ describe("activity content validations", () => {
...
+      });
+
+      it("with multiple attachments but only one is usable/valid, returns the valid one", () => {
```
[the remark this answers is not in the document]
shannonwells: This one tests when there are two attachments but only one is valid.  In that case we should return only the valid one, but you're right I meant to replace this with a strictEqual check, thanks.

src/core/activityContent/validation.test.ts
```diff
@@ -175,17 +293,82 @@ describe("activity content validations", () => {
...
+                },
+              ],
+            },
+          ],
+        };
+
+        expect(requireGetSupportedContentAttachments([validAttachment, invalidAttachment])).toHaveLength(1);
+      });
+
+      it("with multiple link attachments and one is invalid, returns the valid one", () => {
```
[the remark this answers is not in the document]
shannonwells: I know these seem really repetitive, but the code path is slightly different when it's a link attachment versus an Image, Video, or Audio attachment.  I added the attachment type to the other descriptions.

src/core/activityContent/validation.ts
```diff
@@ -27,8 +28,9 @@ import { isRecord, isString, isNumber, isArrayOfType } from "../utilities/valida
  */
 const ISO8601_REGEX = /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(\.\d{1,})?(Z|[+-][01][0-9]:[0-5][0-9])?$/;
 const HREF_REGEX = /^https?:\/\/.+/;
-const DURATION_REGEX = /^-?P[0-9]+Y?([0-9]+M)?([0-9]+D)?(T([0-9]+H)?([0-9]+M)?([0-9]+(\.[0-9]+)?S)?)?$/;
+const DURATION_REGEX = /^-?P([0-9]+Y)?([0-9]+M)?([0-9]+D)?(T([0-9]+H)?([0-9]+M)?([0-9]+(\.[0-9]+)?S)?)?$/;
```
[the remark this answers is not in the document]
shannonwells: Except that regex completely fails on all valid examples I tried.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,388 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
-  if (obj["type"] !== "Link") return false;
-  if (!isString(obj["href"])) return false;
-  if (!isString(obj["mediaType"])) return false;
-  if (!isArrayOfType(obj["hash"], isActivityContentHash)) return false;
-
-  return true;
+const InvalidRecordMessage = (extra?: string): string => {
+  return `${extra || "object"} is not a record type`;
+};
+const requireIsString = (obj: unknown, name: string): string => {
```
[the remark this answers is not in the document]
shannonwells: These need to return the object because they're used to guarantee the type returned in order to be passed to other functions. I'll try to come up with a better name.

src/core/activityContent/validation.ts
shannonwells: Let's discuss what should be in those - I noticed some checks in there that seemed to conflate validation and type checking.  The error-throwing type checks for profile and note are quite a bit simpler now.  We could make non-error throwing versions of those maybe?

src/core/activityContent/validation.test.ts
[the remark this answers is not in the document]
rosalinekarr: No, I think I just misread the test.

src/core/activityContent/validation.test.ts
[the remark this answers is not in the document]
rosalinekarr: in that case, maybe we can change the descriptions to clarify the difference? something like `it("returns valid attachments for Link type attachments")` here?

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
 
-const isActivityContentLink = (obj: unknown): obj is ActivityContentLink => {
-  if (!isRecord(obj)) return false;
-  if (obj["type"] !== "Link") return false;
-  if (!isString(obj["href"])) return false;
-
-  if (obj["name"] && !isString(obj["name"])) return false;
+const requireIsActivityContentTag = (obj: unknown): obj is ActivityContentTag => {
+  if (!isRecord(obj)) throw new InvalidActivityContentError(invalidRecordMessage("ActivityContentTag"));
+  if (obj["type"] && obj["type"] === "Mention") {
```
acruikshank, line 217: nit: The first clause is redundant.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
+    if (obj["published"] && !isValidPublishedField(obj["published"]))
+      throw new InvalidActivityContentError("ActivityContentProfile published is invalid");
+    if (obj["icon"]) {
+      for (const icon of obj["icon"]) {
+        requireToActivityContentLink(icon);
       }
-    } else {
-      if (!isValidActivityContentLocation(obj["location"])) return false;
     }
+    if (obj["location"]) requireIsActivityContentLocation(obj["location"]);
```
acruikshank, line 434: is it no longer valid for a profile to have multiple locations? That should be mentioned in changelog if so.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
+  if (requireIsActivityContentNoteType(obj)) {
+    if (obj["tag"]) requireIsActivityContentTag(obj["tag"]);
+
+    if (obj["published"] && !isValidPublishedField(obj["published"]))
+      throw new InvalidActivityContentError("ActivityContentNote published is invalid");
+    if (obj["attachment"]) {
+      const attachments = requireToArray(obj["attachment"], "ActivityContentNote attachment");
+      requireGetSupportedContentAttachments(attachments);
     }
+    if (obj["location"]) requireIsActivityContentLocation(obj["location"]);
```
acruikshank, line 405: Are multiple locations not supported for notes anymore? If so that should be noted in changelog.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
 
+  requireToString(obj["algorithm"], "ActivityContentHash algorithm");
   return true;
 };
 
-const isActivityContentVideo = (obj: unknown): obj is ActivityContentVideo => {
-  if (!isRecord(obj)) return false;
-  if (obj["type"] !== "Video") return false;
-  if (!isArrayOfType(obj["url"], isActivityContentVideoLink)) return false;
+const requireIsActivityContentLocation = (obj: unknown): obj is ActivityContentLocation => {
```
acruikshank, line 200: This validation appears to have loosened quite a bit (i.e. no restrictions on units or bounds on values). Is that ok? The previous version seems dangerously strict to me.

acruikshank (approved):
Overall it looks like great work. Though this is mostly a refactor, some of the validation rules have changed in the process. Validation changes are always likely to trip up downstream code, so it would be good to document this as specifically as possible in the changelog. I've noted all the instances I could find.

src/core/activityContent/validation.test.ts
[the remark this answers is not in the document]
shannonwells: I realized I should not have put the test cases for a single attachment with multiple URLs in the "multiple attachments."  test suite. This is going to be a source of confusion I think for everyone.  Anyway I  moved them into their proper test suites and adjusted descriptions.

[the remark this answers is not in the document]
shannonwells: And also in that process I discovered a new error case and added a couple of tests for that. So if there is a single attachment which fails the type checks, it throws regardless and the attachment is invalid.  Also I added some clarifying examples in `requireGetSupportedContentAttachments`  as to when one should expect something to throw vs. returning validated attachments.   This seems like it would be the most desirable behavior to me so lmk if you disagree.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
 
   return true;
 };
 
-const isActivityContentMention = (obj: unknown): obj is ActivityContentMention => {
-  if (!isRecord(obj)) return false;
-  if (obj["type"] !== "Mention") return false;
-  if (!isDSNPUserId(obj["id"])) return false;
+const requireIsActivityContentAudioLink = (obj: unknown): obj is ActivityContentAudioLink => {
+  if (!isRecord(obj)) throw new InvalidActivityContentError(invalidRecordMessage("ActivityContentAudioLink"));
```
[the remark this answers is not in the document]
shannonwells: You can't move these because of how TypeScript checking is done.

src/core/activityContent/validation.ts
```diff
@@ -40,386 +42,398 @@ const SUPPORTED_IMAGE_MEDIA_TYPES = [
...
-  if (obj["type"] !== "Image") return false;
-  if (!isArrayOfType(obj["url"], isActivityContentImageLink)) return false;
+  return { type: "Link", href: "", ...obj };
+};
 
-  if (obj["name"] && !isString(obj["name"])) return false;
+const requireIsActivityContentHash = (obj: unknown): obj is ActivityContentHash => {
+  if (!isRecord(obj)) throw new InvalidActivityContentError(invalidRecordMessage("ActivityContentHash"));
+  const hashVal = requireToString(obj["value"], "ActivityContentHash value field");
+  if (!hashVal.match(HASH_REGEX)) throw new InvalidActivityContentError("ActivityContentHash value is invalid");
```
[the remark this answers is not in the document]
shannonwells: The hash checks you've tagged were already there; they've just been refactored to throw now.  But I'll note all of this in the Changelog.

src/core/activityContent/validation.ts
shannonwells: No, we discussed it and decided while you were out.  I'll put it in the Changelog.

shannonwells:
@acruikshank , @rosalinekarr Please have a look at the Changelog entry and let me know if you think it's sufficient.

src/core/activityContent/validation.ts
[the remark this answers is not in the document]
rosalinekarr: I'm really surprised the W3C spec is incorrect. Is there somewhere we can report the mistake to them?

src/core/activityContent/validation.ts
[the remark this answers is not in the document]
rosalinekarr: A type guard function can also guarantee the type: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

src/core/activityContent/validation.ts
rosalinekarr: yep, `isActivityContentNote` and `isActivityContentProfile` could just check that all the appropriate keys are present and have the correct type like they did before. they don't need to do regexes or anything.

rosalinekarr (approved):
only non-blocking feedback left

src/core/activityContent/validation.ts
shannonwells: There's a lot of potential range on location data, and both decimal and negative values are valid. I think those kinds of checks should be left to consumers of that optional data.

PR #133, opened 2021-08-12
Ensure we fetch registrations using Ethereum checksum address

shannonwells:
Problem
=======
There was a bug whereby if we passed in the regular Ethereum address, fetching registrations would fail because it's expecting the checksummed version. 

Kind of [part of #178933068](https://www.pivotaltracker.com/story/show/178933068)

Solution
========
Call ethers.utils.getAddress to ensure we use checksum address version in `getRegistrationsByWalletAddress` to ensure we provide the correct version of the address.
with @wilwade , @acruikshank 

Double Checks:
---------------
- [x] N/A Any new modules need to be exported?
- [x] N/A Are new methods in the right module?
- [x] N/A Are new methods/modules in core or not (i.e. porcelain or plumbing)?
- [x] N/A Do you have good documentation on exported methods?

Change summary:
---------------
* Ensure we fetch registrations using Ethereum checksum address in getRegistrationsByWalletAddress
* Update changelog

Steps to Verify:
----------------
Tests should all pass.
If you run a local hardhat node and deploy contracts, the following test code output the registration twice:
```javascript
import {ethers} from "ethers";
import * as sdk from "@dsnp/sdk";

const main = async () => {
  const providerUrl = "http://localhost:8545"

  const account = "0xa0ee7a142d267c1f36714e4a8f75612f20a79720"
  const accountAlt = "0xa0Ee7A142d267C1f36714E4a8F75612F20a79720"

  const pk = "0xdf57089febbacf7ba0bc227dafbffa9fc08a93fdc68e1e42411a14efcf23656e"
  const provider = new ethers.providers.JsonRpcProvider(providerUrl);
  const signer = new ethers.Wallet(pk, provider);


  sdk.setConfig({provider, signer})
  const res = await sdk.handles.createRegistration(account, "tomasina")
  console.log({res})

  const result = await sdk.core.contracts.registry.getRegistrationsByWalletAddress(account)
  console.log({result})

  const resultAlt = await sdk.core.contracts.registry.getRegistrationsByWalletAddress(accountAlt)
  console.log({resultAlt})
}

main();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+Fix bug in `getRegistrationsByWalletAddress` by normalizing address to checksum version 
 
 ## [2.0.2] - 2021-08-11
 ### Added
--- a/src/core/contracts/identity.ts
+++ b/src/core/contracts/identity.ts
@@ -422,7 +422,7 @@ export const resolveDelegatesFor = async (
 
   const filteredDelegateLogData = lastLogForIdentityAddress.filter((delegate: DelegateLogData) => {
     return (
-      delegate.delegate == ethereumAddress &&
+      delegate.delegate.toLowerCase() == ethereumAddress.toLowerCase() &&
       (isDelegateAddLogData(delegate) || (isDelegateRemoveLogData(delegate) && !isRemoved(delegate)))
     );
   });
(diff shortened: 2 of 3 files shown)


rosalinekarr (approved):

amparks100 (approved):

wilwade (changes requested):
I'm a bit confused. I thought the bug was the comparison wasn't correct in contracts/identity.ts:425

While getting the canonical address is maybe good, I don't think we can know for sure that the log data is also canonical.

rosalinekarr:
@shannonwells, just an fyi, this is the bug I originally open for this when we found the problem: https://www.pivotaltracker.com/story/show/179211543

I went ahead and assigned it to you and marked it as finished since this PR is up for review

wilwade (approved):

PR #143, opened 2021-08-19
Features/read file conversions 178901912

rosalinekarr:
Problem
=======
The batch function `readFile` currently passes records directly from Parquet to the callback, meaning almost all fields are Uint8arrays instead of parsed and formatted announcements.
[#178901912](https://www.pivotaltracker.com/story/show/178901912)

Solution
========
Update `readFile` to parse announcements from records before passing them to the callback.

Change summary:
---------------
* Updated `readFile` to pass parsed announcements
* Added tests for `readFile`
* Changed `createdAt` in announcements to be a `BigInt` instead of a `number`
* Moved some misplaced tests from `activityContent/factories.test.ts` to `announcements/factories.test.ts`


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ## [2.1.1] - 2021-08-17
 ### Changed
 - Updated dependencies
+- Changed batch.readFile() to pass parsed announcements instead of parquet records to callback
+- Changed createdAt on announcements to BigInt instead of number
```
acruikshank, line 9: It would be good to highlight these are breaking changes.

acruikshank (approved):
This looks good. These are definitely breaking changes, so it definitely needs to go in to the 3.0 branch.

CHANGELOG.md
wilwade: Also needs to be under Unreleased.

PR #150, opened 2021-08-23
Better Typed Announcements

wilwade:
Problem
=======
While working on updating the SDK to match spec v0.10.0, I found some type issues that could be resolved by
- Change AnnouncementWithSignature to take an AnnouncementType instead of a TypedAnnouncement
- use bigint for types instead of BigInt

Solution
========
- Change AnnouncementWithSignature to take an AnnouncementType instead of a TypedAnnouncement
- use bigint for types instead of BigInt

NOTE: No exported types

Double Checks:
---------------

### DIFF
--- a/src/core/announcements/factories.ts
+++ b/src/core/announcements/factories.ts
@@ -22,19 +22,53 @@ export type Announcement = TypedAnnouncement<AnnouncementType>;
 /**
  * TypedAnnouncement: an Announcement with a particular AnnouncementType
  */
-export interface TypedAnnouncement<T extends AnnouncementType> {
+export type TypedAnnouncement<T extends AnnouncementType> = {
   announcementType: T;
   fromId: DSNPUserId;
-  createdAt: BigInt;
-}
+  createdAt: bigint;
+} & (TombstoneFields | BroadcastFields | ReplyFields | ReactionFields | GraphChangeFields | ProfileFields);
+
+type TombstoneFields = {
+  announcementType: AnnouncementType.Tombstone;
+  targetAnnouncementType: AnnouncementType;
+  targetSignature: HexString;
+};
+
+type BroadcastFields = {
+  announcementType: AnnouncementType.Broadcast;
+  contentHash: HexString;
+  url: string;
+};
+
+type ReplyFields = {
+  announcementType: AnnouncementType.Reply;
+  contentHash: HexString;
+  inReplyTo: DSNPAnnouncementURI;
+  url: string;
+};
+
+type ReactionFields = {
+  announcementType: AnnouncementType.Reaction;
+  emoji: string;
+  inReplyTo: DSNPAnnouncementURI;
+};
+
+type GraphChangeFields = {
+  announcementType: AnnouncementType.GraphChange;
+  changeType: DSNPGraphChangeType;
+  objectId: DSNPUserId;
+};
+
+type ProfileFields = {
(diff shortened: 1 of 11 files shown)


src/core/batch/batch.ts
```diff
@@ -27,25 +27,23 @@ interface BatchFileData {
...
 
 const parseAnnouncement = <T extends SignedAnnouncement>(record: ParquetRecord): T => {
-  const schema = getSchemaFor(Number(record.announcementType));
-  const announcement: Record<string, string | number | BigInt> = {};
+  const schema = getSchemaFor(record.announcementType as AnnouncementType);
 
-  for (const key in schema) {
-    if (schema[key].type === "BYTE_ARRAY") {
-      announcement[key] = record[key].toString();
+  const announcement = Object.entries(schema).reduce<Record<string, string | number | bigint>>((acc, [key, s]) => {
```
wilwade, line 37: The types really didn't like working with schema and announcement in this way, so I switched over to a reduce setup.

src/core/batch/batch.ts
```diff
@@ -27,25 +27,26 @@ interface BatchFileData {
...
-      announcement[key] = record[key].toString();
-    } else if (typeof record[key] === "number") {
-      announcement[key] = Number(record[key]);
-    } else if (typeof record[key] === "bigint") {
-      announcement[key] = BigInt(record[key].toString());
-    }
-  }
+  const schemas = getSchemaFor(record.announcementType as AnnouncementType);
+
+  const announcement = Object.entries(schemas).reduce<Record<string, string | number | bigint>>(
```
wilwade, line 37: The types really didn't like working with schema and announcement in this way, so I switched over to a reduce setup.

src/core/announcements/validation.test.ts
```diff
@@ -93,7 +93,8 @@ describe("validation", () => {
           createdAt: BigInt(+Date.now()),
           objectId: "not a valid id",
         };
-        const signedAnnouncement = await sign(announcement);
+        /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+        const signedAnnouncement = await sign(announcement as any);
```
rosalinekarr, line 97: why do we need this `any`? can't we cast the invalid announcement to `SignedGraphChangeAnnouncement`?

src/core/announcements/factories.ts
```diff
@@ -22,19 +22,53 @@ export type Announcement = TypedAnnouncement<AnnouncementType>;
...
+  announcementType: AnnouncementType.GraphChange;
+  changeType: DSNPGraphChangeType;
+  objectId: DSNPUserId;
+};
+
+type ProfileFields = {
+  announcementType: AnnouncementType.Profile;
+  contentHash: HexString;
+  url: string;
+};
```
rosalinekarr, line 66: why is this preferable over the extended interfaces?

rosalinekarr (approved):
i have a couple questions, but otherwise looks good

src/core/announcements/factories.ts
wilwade: It let's the type resolve from just the AnnouncementType generic correctly. The interface extensions are weaker and wouldn't be able to resolve the fields properly.

src/core/announcements/validation.test.ts
wilwade: Good idea

PR #159, opened 2021-08-24
Updates for Spec v0.10.0

wilwade:
Problem
=======
SDK needed to be updated to the new version of the spec
[#179299786](https://www.pivotaltracker.com/story/show/179299786)

Solution
========
Updated serialization for decimal and hexadecimal values and Parquet Schema changes

Double Checks:
---------------

Change summary:
---------------
- Use BigInt for DSNPUserId
- Store byte data in parquet as bytes
- Parquet Schema changes
- Use Spec v0.10.0 serialization for signatures


src/core/announcements/serialization.ts
```diff
@@ -17,50 +14,9 @@ export const serialize = (announcement: Announcement): string => {
...
 
-/**
- * serializeToHex() takes in just about anything and attempts to coerce the value to DSNP hexadecimal standard
- *
- * @param value - DSNP User Id/URI BigNumber BigInt number and even unknown
- * @throws {@link InvalidHexadecimalSerialization}
- * Thrown if the coercion failed
- * @returns A hexadecimal serialized value
- */
-export const serializeToHex = (
```
wilwade, line 28: This is one of the big changes. The hex serialization is no longer needed as we are not trying to serialize a wide range of values and strip leading zeros.

This is mostly replaced not by `serializeValue`, but `convertToDSNPUserId` and `convertToDSNPUserURI` from #158

src/core/batch/batch.ts
```diff
@@ -51,6 +51,21 @@ const parseAnnouncement = <T extends SignedAnnouncement>(record: ParquetRecord):
   return announcement as unknown as T;
 };
 
+const encodeParquetRecord = <T extends AnnouncementType>(announcement: AnnouncementWithSignature<T>): ParquetRecord => {
```
wilwade, line 54: This makes sure the values going into Parquet are correct for parquetjs.

Aka most of it is supported out of the box, but hex as byte arrays is not. So this translates those hex values into Uint8Arrays which are supported. `parseAnnouncement` goes the other way back out.

src/core/batch/buffers.ts
```diff
@@ -0,0 +1,38 @@
+import { HexString } from "../../types/Strings";
+
```
wilwade, line 2: The `uint8ArrayToHex` is from https://blog.xaymar.com/2020/12/08/fastest-uint8array-to-hex-string-conversion-in-javascript/ and the `hexToUint8Array` is just reversing it.

src/core/identifiers/identifiers.ts
```diff
@@ -17,15 +19,15 @@ export type DSNPAnnouncementURI = string;
...
+  return id.match(/^dsnp:\/\/[0-9]{1,20}\/0x[0-9a-f]{64}$/) !== null;
 };
 
 /**
  * DSNPUserId represents a DSNP user id following the DSNP
  * [Identifiers](https://github.com/LibertyDSNP/spec/blob/main/pages/Identifiers.md)
  * specification.
  */
-export type DSNPUserId = string;
+export type DSNPUserId = bigint;
```
wilwade, line 30: Important change!

enddynayn (approved):

src/core/batch/buffers.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param arr - The array or buffer to translate to hexadecimal
+ * @returns The hexadecimal string with a 0x prefix
+ */
+export const uint8ArrayToHex = (arr: Uint8Array | Buffer): HexString => {
+  let out = "";
+  for (let idx = 0, edx = arr.length; idx < edx; idx++) {
+    out += LUT_HEX_8B[arr[idx]];
+  }
+  return `0x${out}`;
+};
```
rosalinekarr, line 38: I wonder if these functions will ever be used elsewhere, and if so, should we move them to the utilities module instead?

rosalinekarr (approved):
One non-blocking thought, but otherwise LGTM

src/core/batch/buffers.ts
wilwade: I would be surprised if they were?

I intentionally didn't export them from the batch module so they couldn't be used by normal sdk users.

PR #167, opened 2021-09-01
fix(subscribeToEvent): allow fromBlock zero value

enddynayn:
Ensure that when setting fromBlock to zero
that is does not get interpreted as falsy
and that it fetches event data from zero block.


aneyzberg (approved):
👍

PR #171, opened 2021-09-14
Config DSNP start block

wilwade:
Problem
=======
As a user of the SDK on a long running network, I don't want to search through millions of blocks I don't care about to get the data. I also don't want to have to manually set the default everywhere I call. I just want a config variable to do that work for me.
[PT #179564954](https://www.pivotaltracker.com/story/show/179564954)

Solution
========
Added an optional config param `dsnpStartBlockNumber` that will get used as the default as genesis instead of the chain genesis.

Double Checks:
---------------

Change summary:
---------------
* Add optional config for `dsnpStartBlockNumber`
* Update methods to use the new `dsnpStartBlockNumber`
* Update `transactionRejectsWith` type which was broken for some IDEs

Also added an example and suggested values in https://github.com/LibertyDSNP/sdk-ts/pull/172

### DIFF
--- a/src/core/contracts/subscription.ts
+++ b/src/core/contracts/subscription.ts
@@ -52,7 +52,7 @@ export type BatchPublicationCallback = (doReceivePublication: BatchPublicationLo
  * @throws {@link MissingProviderConfigError}
  * Thrown if the provider is not configured.
  * @param doReceivePublication - The callback function to be called when an event is received
- * @param filter - Any filter options for including or excluding certain events
+ * @param filter - Any filter options for including or excluding certain events (supports fromBlock: "latest" | "dsnp-start-block" | number)
  * @returns A function that can be called to remove listener for this type of event
  */
 export const subscribeToBatchPublications = async (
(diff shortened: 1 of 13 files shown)


src/test/customMatchers.ts
```diff
@@ -1,15 +1,6 @@
 import { ContractTransaction } from "ethers";
 import { getVmError } from "../core/contracts/contract";
 
-declare global {
```
wilwade, line 4: Apparently the better way to do this is to add it in a jest.d.ts file and not here.

src/core/contracts/subscription.ts
```diff
@@ -35,7 +35,7 @@ export interface ParsedLog {
 
 export interface BatchFilterOptions {
   announcementType?: number;
-  fromBlock?: number;
+  fromBlock?: FromBlockNumber;
```
enddynayn, line 38: nice

src/core/contracts/utilities.ts
```diff
@@ -86,23 +87,49 @@ export interface LogEventData {
...
+ * Convert a user block value to an actual block number.
+ * Defaults to "latest" for subscription needs
+ *
+ * @param userFromBlock - undefined results in using defaultZeroOrLatest
+ * @param defaultZeroOrLatest - undefined = latest or 0?
+ * @returns A block number or "latest"
+ */
+export const getFromBlockDefault = (
+  userFromBlock: FromBlockNumber | undefined,
+  defaultZeroOrLatest: 0 | "latest"
```
enddynayn, line 108: nice

enddynayn (approved):

PR #175, opened 2021-09-22
Feature/fetch logs paged

shannonwells:
Problem
=======
We want the ability to fetch Batch Publications asychronously and lazily loaded.
[Finishes #179436295](https://www.pivotaltracker.com/story/show/179436295)

Solution
========
* Create a class AsyncPublicationsIterator that conforms to AsyncIterator (but only in that it has an async next() function)
* Create a function that returns the new iterator after doing some range checking.
* Tests for same

How `next()` works:
1. Calling `next()` on the iterator makes it search for filtered logs on chain, `walkbackBlocks` at a time until it gets results.  
2. Once it gets at least one result it, returns an "IteratorResult" which is an object containing a "done" boolean meaning no more results are left, and the next Publication result, or undefined if there are none.
3. Subsequent calls to `next()` advance the index into the fetched publications until it runs out.  
4. When next has returned the last of the fetched results, calling `next()` repeats the process of walking back until it gets results, or until the `earliestBlock` value has been reached, whichever comes first.

Double Checks:
---------------
- [X] Any new modules need to be exported? N/A
- [X] Are new methods in the right module?  I think so?
- [X] Are new methods/modules in core or not (i.e. porcelain or plumbing)?
- [X] Do you have good documentation on exported methods?

Change summary:
---------------
* New class, AsyncPublicationsIterator
* Creator function for above
* Tests for same

Steps to Verify:
----------------
1. Tests should all pass.
1. You should be able to create and use the AsyncPublicationsIterator when you import this version of the sdk.

### DIFF
--- a/src/core/contracts/subscription.ts
+++ b/src/core/contracts/subscription.ts
@@ -1,17 +1,14 @@
 import { HexString } from "../../types/Strings";
 import { ethers } from "ethers";
 import { ConfigOpts, requireGetProvider } from "../config";
-import { dsnpBatchFilter, Publication } from "./publisher";
+import { dsnpBatchFilter } from "./publisher";
 import { Publisher__factory, Registry } from "../../types/typechain";
 import { LogDescription } from "@ethersproject/abi";
 import { FromBlockNumber, getFromBlockDefault, LogEventData, subscribeToEvent, UnsubscribeFunction } from "./utilities";
 import { getContract, RegistryUpdateLogData } from "./registry";
 import { convertToDSNPUserURI } from "../identifiers";
-import { EventFilter } from "ethers/lib/ethers";
-import { Log as EventLog } from "@ethersproject/abstract-provider";
 
 const PUBLISHER_DECODER = new ethers.utils.Interface(Publisher__factory.abi);
-
 /**
  * BatchPublication represents a struct for a publication data
  */
@@ -72,7 +69,7 @@ export const subscribeToBatchPublications = async (
   return subscribeToEvent(provider, batchFilter, doReceiveEvent, getFromBlockDefault(filter?.fromBlock, "latest"));
 };
 
-const decodeLogsForBatchPublication = (logs: ethers.providers.Log[]): BatchPublicationLogData[] => {
+export const decodeLogsForBatchPublication = (logs: ethers.providers.Log[]): BatchPublicationLogData[] => {
   return logs
     .map((log: ethers.providers.Log) => {
       try {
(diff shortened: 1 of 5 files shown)


src/core/contracts/utilities.ts
```diff
@@ -181,3 +186,148 @@ export const subscribeToEvent = async (
...
+ * BlockRangeOptions: represents parameters for specifying chain filter and block limits.
+ */
+export type BlockRangeOptions = {
+  filter: EventFilter;
+  walkbackBlocks: number;
+  latestBlock: number;
+  earliestBlock: number;
+};
+
+export const MAX_ITERATOR_WALKBACK_BLOCKS = 10000;
```
shannonwells, line 200: I'm unsure about this constant even being here, however, I figured it would be good to set some kind of maximum, and  this is the same value as the current Ethereum log fetch limit.

src/core/contracts/utilities.ts
```diff
@@ -181,3 +186,148 @@ export const subscribeToEvent = async (
...
+  requireValidWalkback(walkbackBlocks);
+  requireValidBlockRange(newestBlock, oldestBlock);
+  const provider = requireGetProvider(opts);
+  const newest = newestBlock || (await provider.getBlockNumber());
+  const oldest = oldestBlock === undefined ? 0 : oldestBlock;
+  return new AsyncPublicationsIterator(
+    {
+      earliestBlock: oldest,
+      latestBlock: newest,
+      walkbackBlocks: Math.min(newest - oldest + 1, walkbackBlocks),
```
shannonwells, line 328: +1 so it's inclusive

src/core/contracts/subscription.ts
```diff
@@ -71,7 +71,7 @@ export const subscribeToBatchPublications = async (
   return subscribeToEvent(provider, batchFilter, doReceiveEvent, getFromBlockDefault(filter?.fromBlock, "latest"));
 };
 
-const decodeLogsForBatchPublication = (logs: ethers.providers.Log[]): BatchPublicationLogData[] => {
+export const decodeLogsForBatchPublication = (logs: ethers.providers.Log[]): BatchPublicationLogData[] => {
```
wilwade, line 74: We should likely doc this if we are going to export it.

src/core/contracts/utilities.ts
```diff
@@ -181,3 +185,148 @@ export const subscribeToEvent = async (
...
+   * \}
+   */
+  public async next(): Promise<IteratorResult<BatchPublicationLogData>> {
+    if (this.shouldFetchLogs()) {
+      await this.fetchUntilGetLogsOrDone();
+    } else {
+      this.logIndex++;
+    }
+    return Promise.resolve({ done: this.doneFetching(), value: this.publications[this.logIndex] });
+  }
```
wilwade, line 279: We need this to support `for await (const nextResult of iterator)`

```suggestion
  }
  
  public [Symbol.asyncIterator](): AsyncIterableIterator<BatchPublicationLogData> {
    return this;
  }
```

wilwade (changes requested):
Few small things to doc or add support for for await, but looks good.

shannonwells:
@wilwade I think I got everything.

wilwade (approved):

PR #184, opened 2022-01-21
Changed license placeholder values

waylandli:
Problem
=======
License place holder value was not changed
Solution
========
Changed the placeholder values to the correct value


wilwade (approved):
👍

